PR #20, opened 2017-04-27
small fixes

usakhelo:
Hi,
this fixes changing diffuse colors from the multi-material's color swatches, makes glass material transparent in viewports and also "fixes" duplicated textures error. 
"Fixes" duplicated textures in the sense that it avoids adding textures with the same name, but in max it is possible to have different texture with the same name. 
It could be possible to track max's map instances in the same way as materials are collected. But textures are created from each material class and I didn't know if I should add another map for the textures to the populate_assembly and pass it to each material.

### DIFF
--- a/src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
+++ b/src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
@@ -677,6 +677,12 @@ void AppleseedDisneyMtl::SetAmbient(Color c, TimeValue t)
 
 void AppleseedDisneyMtl::SetDiffuse(Color c, TimeValue t)
 {
+    Point3 p;
+    Interval iv;
+    m_pblock->SetValue( ParamIdBaseColor, t, Point3(c.r,c.g,c.b));
+
+    m_pblock->GetValue( ParamIdBaseColor, t, p, iv );
+    m_base_color = Color(p.x,p.y,p.z);
 }
 
 void AppleseedDisneyMtl::SetSpecular(Color c, TimeValue t)
--- a/src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
+++ b/src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
@@ -677,12 +677,12 @@ void AppleseedDisneyMtl::SetAmbient(Color c, TimeValue t)
 
 void AppleseedDisneyMtl::SetDiffuse(Color c, TimeValue t)
 {
-    Point3 p;
+    Color nc;
     Interval iv;
-    m_pblock->SetValue( ParamIdBaseColor, t, Point3(c.r,c.g,c.b));
+    m_pblock->SetValue(ParamIdBaseColor, t, c);
 
-    m_pblock->GetValue( ParamIdBaseColor, t, p, iv );
-    m_base_color = Color(p.x,p.y,p.z);
+    m_pblock->GetValue(ParamIdBaseColor, t, nc, iv );
+    m_base_color = nc;
 }
 
 void AppleseedDisneyMtl::SetSpecular(Color c, TimeValue t)
--- a/src/appleseed-max-impl/appleseedglassmtl/appleseedglassmtl.cpp
+++ b/src/appleseed-max-impl/appleseedglassmtl/appleseedglassmtl.cpp
@@ -54,6 +54,7 @@
 #include <iparamm2.h>
 #include <stdmat.h>
 #include <strclass.h>
+#include <hsv.h>
 
 // Windows headers.
 #include <tchar.h>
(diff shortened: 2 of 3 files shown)


src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
```diff
@@ -677,6 +677,12 @@ void AppleseedDisneyMtl::SetAmbient(Color c, TimeValue t)
 
 void AppleseedDisneyMtl::SetDiffuse(Color c, TimeValue t)
 {
+    Point3 p;
+    Interval iv;
+    m_pblock->SetValue( ParamIdBaseColor, t, Point3(c.r,c.g,c.b));
```
dictoon: Remove space after `(`.
Add spaces after commas.

src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
```diff
@@ -677,6 +677,12 @@ void AppleseedDisneyMtl::SetAmbient(Color c, TimeValue t)
...
+
+    m_pblock->GetValue( ParamIdBaseColor, t, p, iv );
```
dictoon: Remove spaces after `(` and before `)`.

src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
dictoon: In fact, isn't there a `SetValue()` method that takes a `Color`?

src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
```diff
@@ -677,6 +677,12 @@ void AppleseedDisneyMtl::SetAmbient(Color c, TimeValue t)
...
+    m_base_color = Color(p.x,p.y,p.z);
```
dictoon: I'd suggest to use `to_color3f()` defined in `utilities.h`.

src/appleseed-max-impl/appleseedglassmtl/appleseedglassmtl.cpp
```diff
@@ -54,6 +54,7 @@
 #include <iparamm2.h>
 #include <stdmat.h>
 #include <strclass.h>
+#include <hsv.h>
```
dictoon: Keep header files sorted alphabetically.

src/appleseed-max-impl/appleseedglassmtl/appleseedglassmtl.cpp
```diff
@@ -591,7 +592,8 @@ float AppleseedGlassMtl::GetShinStr(int mtlNum, BOOL backFace)
 
 float AppleseedGlassMtl::GetXParency(int mtlNum, BOOL backFace)
 {
-    return 0.0f;
+    Color hsv = RGBtoHSV (m_refraction_tint);
```
dictoon: Make const.

src/appleseed-max-impl/utilities.cpp
```diff
@@ -171,18 +171,23 @@ std::string insert_texture_and_instance(
     }
 
     const std::string texture_name = wide_to_utf8(bitmap_tex->GetName());
-    base_group.textures().insert(
-        asr::DiskTexture2dFactory::static_create(
+    if (base_group.textures().get_by_name(texture_name.c_str()) == nullptr)
```
dictoon: As you hinted in your PR's description, doesn't this create a bug when two 3ds Max maps have the same name? If so, we'll need to track maps properly, I'm afraid.

In any case we can keep these changes for now.

src/appleseed-max-impl/utilities.cpp
```diff
@@ -171,18 +171,23 @@ std::string insert_texture_and_instance(
...
+    if (base_group.texture_instances().get_by_name(texture_instance_name.c_str()) == nullptr)
+    {
+        base_group.texture_instances().insert(
+            asr::TextureInstanceFactory::create(
             texture_instance_name.c_str(),
             texture_instance_params,
             texture_name.c_str()));
-
+    }
     return texture_instance_name;
```
dictoon: Add blank line after closing brace (`}`).

dictoon:
By the way, you had initially mentioned a fix about appleseed materials in the Slate editor: I'm very interested about this. What was the problem?

src/appleseed-max-impl/utilities.cpp
usakhelo: This will make renderer use one map even if max's scene contains other maps with the same name. With this change renderer will keep rendering and not disappearing like it does now.

usakhelo:
Yes I mentioned about slate editor but I was wrong. Bug still exists and it is indeed interesting. Looks like slate previews updated differently.

dictoon:
Thanks!

Yeah that Slate editor bug is really puzzling, I unfortunately still don't know enough about Max's SDK to figure it out...

PR #28, opened 2017-06-12
Physical camera and IMtlRender_Compatibility support

usakhelo:
Hi,

I'm trying add Depth of Field and Bokeh support by converting Physical camera's properties. Works only with 3ds max version 2016 and higher. Depth of field and Bokeh is turning on when Enable Depth of Field checkbox is on in Physical Camera properties.
Properties being transferred to appleseed are:
- Aperture
- Target Distance (distance to target or custom distance)
- Circular and Bladed Bokeh
- Blade number and Rotation

Also I've added IMtlRender_Compatibility_MtlBase implementation to each material to hide them from material list when any renderer other than appleseed is chosen in 3ds max.

Thanks for reviewing,
Sergo.

### DIFF
--- a/src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
+++ b/src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
@@ -30,10 +30,11 @@
 #include "appleseeddisneymtl.h"
 
 // appleseed-max headers.
-#include "bump/bumpparammapdlgproc.h"
-#include "bump/resource.h"
 #include "appleseeddisneymtl/datachunks.h"
 #include "appleseeddisneymtl/resource.h"
+#include "appleseedrenderer/appleseedrenderer.h"
+#include "bump/bumpparammapdlgproc.h"
+#include "bump/resource.h"
 #include "main.h"
 #include "utilities.h"
 #include "version.h"
@@ -807,6 +808,11 @@ Bitmap* AppleseedDisneyMtlBrowserEntryInfo::GetEntryThumbnail() const
 // AppleseedDisneyMtlClassDesc class implementation.
 //
 
+AppleseedDisneyMtlClassDesc::AppleseedDisneyMtlClassDesc()
+{
+    IMtlRender_Compatibility_MtlBase::Init(*this);
+}
+
 int AppleseedDisneyMtlClassDesc::IsPublic()
 {
     return TRUE;
@@ -854,4 +860,9 @@ FPInterface* AppleseedDisneyMtlClassDesc::GetInterface(Interface_ID id)
 HINSTANCE AppleseedDisneyMtlClassDesc::HInstance()
 {
     return g_module;
+}
+
+bool AppleseedDisneyMtlClassDesc::IsCompatibleWithRenderer(ClassDesc& renderer_class_desc)
+{
+    return (renderer_class_desc.ClassID() == AppleseedRenderer::get_class_id()) ? true : false;
 }
--- a/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
@@ -75,6 +75,11 @@ AppleseedRendererClassDesc g_appleseed_renderer_classdesc;
 // AppleseedRenderer class implementation.
 //
 
+Class_ID AppleseedRenderer::get_class_id()
+{
+    return AppleseedRendererClassId;
+}
+
 AppleseedRenderer::AppleseedRenderer()
   : m_settings(RendererSettings::defaults())
 {
--- a/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
@@ -401,6 +401,7 @@ int AppleseedRenderer::Render(
         build_project(
             m_entities,
             m_default_lights,
+            m_view_node,
             m_view_params,
             m_rend_params,
             frame_rend_params,
--- a/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.h
+++ b/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.h
@@ -51,6 +51,8 @@ class AppleseedRenderer
   : public Renderer
 {
   public:
+    static Class_ID get_class_id();
+
     AppleseedRenderer();
 
     virtual Class_ID ClassID() override;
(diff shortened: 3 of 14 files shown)


src/appleseed-max-impl/appleseedsssmtl/appleseedsssmtl.cpp
```diff
@@ -828,3 +834,12 @@ HINSTANCE AppleseedSSSMtlClassDesc::HInstance()
 {
     return g_module;
 }
+
+bool AppleseedSSSMtlClassDesc::IsCompatibleWithRenderer(ClassDesc& rendererClassDesc)
+{
+    Class_ID classID = rendererClassDesc.ClassID();
```
dictoon: The whole method could be simplified to

    return renderer_class_desc.ClassID() == AppleseedRenderer::get_class_id();

This applies to other materials as well.

src/appleseed-max-impl/appleseedsssmtl/appleseedsssmtl.cpp
```diff
@@ -828,3 +834,12 @@ HINSTANCE AppleseedSSSMtlClassDesc::HInstance()
 {
     return g_module;
 }
+
+bool AppleseedSSSMtlClassDesc::IsCompatibleWithRenderer(ClassDesc& rendererClassDesc)
```
dictoon: With our naming convention: `renderer_class_desc`.

This applies to other materials as well.

src/appleseed-max-impl/appleseedsssmtl/appleseedsssmtl.h
```diff
@@ -178,8 +179,10 @@ class AppleseedSSSMtlBrowserEntryInfo
 
 class AppleseedSSSMtlClassDesc
   : public ClassDesc2
+  , IMtlRender_Compatibility_MtlBase
```
dictoon: Inherit publicly:

    public IMtlRender_Compatibility_MtlBase

This applies to other materials as well.

src/appleseed-max-impl/appleseedsssmtl/appleseedsssmtl.cpp
```diff
@@ -32,6 +32,7 @@
 // appleseed-max headers.
 #include "appleseedsssmtl/datachunks.h"
 #include "appleseedsssmtl/resource.h"
+#include "appleseedrenderer/appleseedrenderer.h"
```
dictoon: Keep header files ordered alphabetically.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.h
```diff
@@ -35,6 +35,7 @@
 // 3ds Max headers.
 #include <maxtypes.h>
 #include <render.h>
+#include <Scene/IPhysicalCamera.h>
```
dictoon: Include only for 3ds Max 2016 and later.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.h
```diff
@@ -54,6 +55,7 @@ foundation::auto_release_ptr<renderer::Project> build_project(
     const std::vector<DefaultLight>&    default_lights,
     const ViewParams&                   view_params,
     const RendParams&                   rend_params,
+    INode*                              view_node,
```
dictoon: Move `view_node` before `view_params` to be consistent with the argument order you chose for `build_camera()` (which makes sense in my opinion).

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1133,10 +1137,44 @@ namespace
...
-                ? asr::PinholeCameraFactory::static_create("camera", params)
-                : asr::OrthographicCameraFactory::static_create("camera", params);
+        asf::auto_release_ptr<renderer::Camera> camera;
+        if (view_params.projType == PROJ_PARALLEL)
+        {
+            camera = asr::OrthographicCameraFactory::static_create("camera", params);
+        }
+        else
+        {
+        #if MAX_RELEASE == MAX_RELEASE_R16
```
dictoon: It is conventional in C/C++ to not indent `#if` / `#else` lines, I suggest following this convention.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.h
```diff
@@ -36,6 +36,7 @@
 #include "foundation/platform/windows.h"    // include before 3ds Max headers
 
 // 3ds Max headers.
+#include <IMtlRender_Compatibility.h>
```
dictoon: It doesn't look like we need this header either.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -35,6 +35,8 @@
 #include "appleseedrenderer/projectbuilder.h"
 #include "appleseedrenderer/renderercontroller.h"
 #include "appleseedrenderer/tilecallback.h"
+#include "appleseeddisneymtl/appleseeddisneymtl.h"
+#include "appleseedenvmap/appleseedenvmap.h"
```
dictoon: It doesn't look like we need these headers.

src/appleseed-max-impl/appleseedsssmtl/appleseedsssmtl.cpp
usakhelo: I had to changed it to
`return (renderer_class_desc.ClassID() == AppleseedRenderer::get_class_id()) ? true : false;`
because ClassID's operator== returns int and compiler was complaining about converting int to bool

usakhelo:
Hi,
I've rebased and pushed with --force. Hopefully it'll work.

src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
```diff
@@ -855,3 +861,8 @@ HINSTANCE AppleseedDisneyMtlClassDesc::HInstance()
 {
     return g_module;
 }
+
+bool AppleseedDisneyMtlClassDesc::IsCompatibleWithRenderer(ClassDesc& renderer_class_desc)
+{
+    return (renderer_class_desc.ClassID() == AppleseedRenderer::get_class_id()) ? true : false;
```
dictoon: You could simplify even further: the `? true : false` part is not needed since the expression is already of type `bool`. Let's leave it like that for now. I might make that simplification later.

dictoon:
Perfect, thank you!

src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
dictoon: My bad!! 3ds Max 2015 defines the `Class_ID` equality operator as

    int operator==(const Class_ID& cid) const;

so your code was in fact perfectly correct!

PR #33, opened 2017-07-10
Translate Disney material into an OSL network

dictoon:


usakhelo:
after a quick review: I am able to compile it and it renders correctly. I'm yet to understand it fully though. I expected to see osl material in studio, but I'm sure I didn't get it right.

dictoon:
The exported scene has `shaders/` as the only search path. If you want to load the scene in appleseed.cli or appleseed.studio you will need to change that search path to the absolute path to the `shaders/max` directory of your appleseed sandbox. For instance on my Windows machine that would be

    N:\appleseed\appleseed\sandbox\shaders\max

There is still a bit of work to be done here in order to make this fully automatic.

Regarding the PR: are there specific parts you don't understand, or is it more a general feeling of being lost? In any case I suggest discussing this on Slack!

PR #51, opened 2017-08-25
Interactive rendering mode

usakhelo:
Hi!
This is my attempt to implement interactive renderer interface for appleseed.
Most of work is happening in interactivesession and interactiverederer files.
I had to change regular renderer class in order to make 3ds max recognize it as a active shade renderer
I also had to change logtarget.cpp class. I'm not sure if I did it correctly. It caused max to crash on certain cases (cancelling or opening render dialog while rendering) and moving LogEntry call to the UI thread helped to avoid crashes.
I'll add comments to the code lines where appropriate.

UPDATE: Forgot to mention that renderer restarts only on current camera movements.

Thanks for review,
Sergo.


src/appleseed-max-impl/appleseedinteractive/interactivesession.cpp
```diff
@@ -0,0 +1,71 @@
...
+void InteractiveSession::abort_render()
+{
+    m_render_ctrl->set_status(asr::IRendererController::AbortRendering);
+}
+
+void InteractiveSession::reininitialize_render()
+{
+    m_render_ctrl->set_status(asr::IRendererController::ReinitializeRendering);
+}
+
```
usakhelo: this three functions control the renderer through renderer controller

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
```diff
@@ -0,0 +1,53 @@
...
+    asr::IRendererController*       render_controller
+    )
+    : TileCallback(bitmap, 0)
+    , m_bitmap(bitmap)
+    , m_iimanager(iimanager)
+    , m_renderer_ctrl(render_controller)
+{
+}
+
+void InteractiveTileCallback::update_window()
```
usakhelo: this function is called from ui thread. I block the renderer thread until this function completes in order to guarantee m_iimanager still exists. m_ui_promise is class member, I'm not sure if it's ok. It is assigned new promise object on each frame_end.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
```diff
@@ -0,0 +1,53 @@
+
+// Interface header.
+#include "appleseedinteractive/interactivetilecallback.h"
+
+namespace asr = renderer;
+
+#define WM_TRIGGER_CALLBACK WM_USER + 4764
```
usakhelo: this value comes from max sdk docs

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -165,6 +188,22 @@ void AppleseedRenderer::GetPlatformInformation(MSTR& info) const
 
 #endif
```
usakhelo: I had to implemented Clone method in order to make interactive renderer actually start rendering.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -165,6 +188,22 @@ void AppleseedRenderer::GetPlatformInformation(MSTR& info) const
...
+        for (int i = 0; i < num_refs; ++i)
+        {
+            new_rend->ReplaceReference(i, remap.CloneRef(GetReference(i)));
+        }
+        BaseClone(this, new_rend, remap);
+    }
+
+    return new_rend;
+}
+
```
usakhelo: I had to implemented Clone method in order to make interactive renderer actually start rendering.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1316,6 +1241,81 @@ namespace
...
+        asf::Transformd::from_local_to_parent(
+            asf::Matrix4d::make_scaling(asf::Vector3d(settings.m_scale_multiplier)) *
+            to_matrix4d(Inverse(view_params.affineTM))));
+
+    return camera;
+}
+
 asf::auto_release_ptr<asr::Project> build_project(
     const MaxSceneEntities&                 entities,
     const std::vector<DefaultLight>&        default_lights,
```
usakhelo: Just made build_camera visible

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -0,0 +1,484 @@
...
+    {
+        RenderEndProc proc(time);
+        proc.BeginEnumeration();
+
+        for (auto node : nodes)
+            node->EnumRefHierarchy(proc);
+
+        proc.EndEnumeration();
+    }
+
```
usakhelo: code above is mostly duplicate of regular renderer's anonymous namespace

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -0,0 +1,484 @@
...
+void AppleseedIInteractiveRender::update_camera(INode* camera)
+{
+    ViewParams view_params;
+    get_view_params_from_view_node(view_params, camera, m_time);
+
+    m_project->get_scene()->get_active_camera()->transform_sequence().set_transform(
+        (float)m_time,
+        asf::Transformd::from_local_to_parent(to_matrix4d(Inverse(view_params.affineTM))));
+}
```
usakhelo: the method above is called when camera is moved. I'm not sure if it's the correct way to update camera in project.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
```diff
@@ -0,0 +1,53 @@
...
+    m_ui_promise = std::promise<void>();
+    if (m_renderer_ctrl->get_status() == asr::IRendererController::ContinueRendering)
+    {
+        std::future<int> ui_future = m_ui_promise.get_future();
+        post_callback(update_caller, (UINT_PTR)this);
+        ui_future.wait();
+    }
+}
+
+void InteractiveTileCallback::post_callback(void(*funcPtr)(UINT_PTR), UINT_PTR param)
```
dictoon: Get rid of this function, just inline it where needed.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.h
```diff
@@ -0,0 +1,45 @@
...
+{
+  public:
+      InteractiveTileCallback(
+        Bitmap*                         bitmap,
+        IIRenderMgr*                    iimanager,
+        renderer::IRendererController*  render_controller);
+
+    virtual void on_progressive_frame_end(const renderer::Frame* frame) override;
+
+    void update_window();
```
dictoon: Get rid of this method.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
dictoon: Get rid of this method, inline in `update_caller()`.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.h
```diff
@@ -0,0 +1,45 @@
...
+    void post_callback(void(*funcPtr)(UINT_PTR), UINT_PTR param);
+    static void update_caller(UINT_PTR param_ptr);
```
dictoon: Make private.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -52,6 +53,8 @@
 
 // 3ds Max headers.
 #include <assert1.h>
```
dictoon: Order alphabetically.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.h
```diff
@@ -61,6 +65,11 @@ class AppleseedRenderer
 
     virtual void DeleteThis() override;
 
+    // Animatable.
+    virtual void* GetInterface(ULONG id) override;
+
+    AppleseedIInteractiveRender* m_interactive_renderer;
```
dictoon: Make private.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -98,9 +102,28 @@ void AppleseedRenderer::GetClassName(MSTR& s)
 
 void AppleseedRenderer::DeleteThis()
 {
+    if (m_interactive_renderer != nullptr)
```
dictoon: The check is redundant, it's perfectly valid to call `delete` on a `nullptr`.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -165,6 +188,22 @@ void AppleseedRenderer::GetPlatformInformation(MSTR& info) const
 
 #endif
 
+RefTargetHandle	AppleseedRenderer::Clone(RemapDir &remap)
```
dictoon: Remove tabulation. Move `&` to the type's side.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -165,6 +188,22 @@ void AppleseedRenderer::GetPlatformInformation(MSTR& info) const
...
+{
+    AppleseedRenderer* new_rend = static_cast<AppleseedRenderer*>(g_appleseed_renderer_classdesc.Create(false));
+    if (DbgVerify(new_rend != nullptr))
```
dictoon: I don't recommend using `DbgVerify()`. Use `DbgAssert()` instead.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -165,6 +188,22 @@ void AppleseedRenderer::GetPlatformInformation(MSTR& info) const
...
+    {
+        const int num_refs = NumRefs();
+        for (int i = 0; i < num_refs; ++i)
```
dictoon: More compactly:

    for (int i = 0, e = NumRefs(); i < e; ++i)

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -98,9 +102,28 @@ void AppleseedRenderer::GetClassName(MSTR& s)
...
+        delete m_interactive_renderer;
     delete this;
 }
 
+void* AppleseedRenderer::GetInterface(ULONG id)
+{
+    if (id == I_RENDER_ID)
+    {
+        if (m_interactive_renderer == nullptr)
+        {
```
dictoon: You can omit braces for single-line bodies.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -165,6 +188,22 @@ void AppleseedRenderer::GetPlatformInformation(MSTR& info) const
...
 #endif
 
+RefTargetHandle	AppleseedRenderer::Clone(RemapDir &remap)
+{
+    AppleseedRenderer* new_rend = static_cast<AppleseedRenderer*>(g_appleseed_renderer_classdesc.Create(false));
+    if (DbgVerify(new_rend != nullptr))
+    {
+        const int num_refs = NumRefs();
+        for (int i = 0; i < num_refs; ++i)
+        {
```
dictoon: You can omit braces for single-line bodies.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.h
```diff
@@ -82,6 +91,8 @@ class AppleseedRenderer
 
 #endif
 
+    virtual RefTargetHandle	Clone(RemapDir	&remap) override;
```
dictoon: Remove tabs.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.h
```diff
@@ -43,6 +43,7 @@
 #include <vector>
 
 // Forward declarations.
+namespace renderer { class Camera; }
```
dictoon: Align opening braces.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.h
```diff
@@ -62,3 +63,10 @@ foundation::auto_release_ptr<renderer::Project> build_project(
     const RendererSettings&             settings,
     Bitmap*                             bitmap,
     const TimeValue                     time);
+
+foundation::auto_release_ptr<renderer::Camera> build_camera(
+    INode*                  view_node,
```
dictoon: Align arguments with other functions.

src/appleseed-max-impl/logtarget.cpp
```diff
@@ -46,6 +46,34 @@
...
+  {
+      std::vector<std::string> lines;
+      DWORD type;
+  };
+
+  void ui_log_writer(UINT_PTR param_ptr)
+  {
+     message_data* data_ptr = reinterpret_cast<message_data*>(param_ptr);
+
+    for (auto line : data_ptr->lines)
```
dictoon: Fix indentation.

src/appleseed-max-impl/logtarget.cpp
```diff
@@ -46,6 +46,34 @@
 
 namespace asf = foundation;
 
+#define WM_TRIGGER_CALLBACK WM_USER+4764
+
+namespace
+{
+  struct message_data
```
dictoon: Fix indentation.

dictoon: Class names use CamelCase: `MessageData`.

src/appleseed-max-impl/logtarget.cpp
```diff
@@ -46,6 +46,34 @@
 
 namespace asf = foundation;
 
+#define WM_TRIGGER_CALLBACK WM_USER+4764
```
dictoon: Always prefer C++ constants to defines, when possible.

src/appleseed-max-impl/logtarget.cpp
```diff
@@ -46,6 +46,34 @@
...
+      DWORD type;
+  };
+
+  void ui_log_writer(UINT_PTR param_ptr)
+  {
+     message_data* data_ptr = reinterpret_cast<message_data*>(param_ptr);
+
+    for (auto line : data_ptr->lines)
+    {
+      GetCOREInterface()->Log()->LogEntry(
```
dictoon: Where does this write messages, actually? It sounds interesting.

src/appleseed-max-impl/logtarget.cpp
```diff
@@ -71,16 +99,29 @@ void LogTarget::write(
...
+              FALSE,
+              _T("appleseed"),
+              _T("[appleseed] %s"),
+              utf8_to_wide(line).c_str());
+        }
+    }
+    else
+    {
+        message_data* data = new message_data();
+        data->lines.clear();
```
dictoon: You don't need to call `clear()`, the vector always starts empty.

src/appleseed-max-impl/logtarget.cpp
```diff
@@ -71,16 +99,29 @@ void LogTarget::write(
         break;
     }
 
-    std::vector<std::string> lines;
-    asf::split(message, "\n", lines);
+    auto is_ui_thread = GetCOREInterface15()->GetMainThreadID() == GetCurrentThreadId();
 
-    for (auto line : lines)
+    if (is_ui_thread)
```
dictoon: What's the logic here? Why do we sometimes write to the log file, sometimes somewhere else?

usakhelo: Hi Franz,
I'm yet to fix all your comments, but I wanted to answer this first - so, I had to change this so that it only writes to log from main thread. Otherwise it was crashing during interactive session, since it run from interactive session thread.
I guess during production render this part always runs from main thread.

src/appleseed-max-impl/logtarget.cpp
usakhelo: It writes to Max.log file in %localappdata%\Autodesk\3dsMax\2016 - 64bit\ENU\Network
Is this what you are asking?

src/appleseed-max-impl/logtarget.cpp
```diff
@@ -71,16 +99,29 @@ void LogTarget::write(
...
     {
-        GetCOREInterface()->Log()->LogEntry(
-            type,
-            FALSE,
-            _T("appleseed"),
-            _T("[appleseed] %s"),
-            utf8_to_wide(line).c_str());
+        std::vector<std::string> lines;
+        asf::split(message, "\n", lines);
+        for (auto line : lines)
```
dictoon: This loop is the same in both cases, what about factorizing it into a separate function?

src/appleseed-max-impl/logtarget.cpp
dictoon: Got it, it's obvious now, I was just confused temporarily :)

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+        m_render_session->start_render();
+    }
+}
+
+void AppleseedIInteractiveRender::EndSession()
+{
+    m_currently_rendering = false;
+
+    if (m_render_session != nullptr)
```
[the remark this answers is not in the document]
usakhelo: EndSession is actually called many times (from the destructor) without any session being created, so it is possible to have m_render_session nullptr here. In BeginSession check was redundant because it is only called when renderering started.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+const Box2& AppleseedIInteractiveRender::GetRegion() const
+{
+    return m_region;
+}
+
+void AppleseedIInteractiveRender::SetDefaultLights(DefaultLight* pDefLights, int numDefLights)
+{
+    m_default_lights.clear();
+    if (pDefLights != nullptr)
```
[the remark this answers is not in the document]
usakhelo: I'm not sure, this is per documentation or example. Some things don't make sense, so I just copied it.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -0,0 +1,484 @@
...
+}
+
+bool AppleseedIInteractiveRender::GetScreenBBox(Box2& sBBox, INode * pINode)
+{
+    return FALSE;
+}
+
+ActionTableId AppleseedIInteractiveRender::GetActionTableId()
+{
+    return NULL;
```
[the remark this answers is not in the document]
usakhelo: Other place can be nullptr while this can't - ActionTableId is DWORD and it complains about nullptr being wrong type.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
```diff
@@ -0,0 +1,53 @@
...
+namespace asr = renderer;
+
+#define WM_TRIGGER_CALLBACK WM_USER + 4764
+
+InteractiveTileCallback::InteractiveTileCallback(
+    Bitmap*                         bitmap,
+    IIRenderMgr*                    iimanager,
+    asr::IRendererController*       render_controller
+    )
+    : TileCallback(bitmap, 0)
```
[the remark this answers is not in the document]
usakhelo: Should I add dummy member variable?

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.h
usakhelo: I don't think I can make  update_caller private, it gets called from UI thread message loop.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -0,0 +1,484 @@
...
+    return project;
+}
+
+void AppleseedIInteractiveRender::update_camera(INode* camera)
+{
+    ViewParams view_params;
+    get_view_params_from_view_node(view_params, camera, m_time);
+
+    m_project->get_scene()->get_active_camera()->transform_sequence().set_transform(
+        (float)m_time,
```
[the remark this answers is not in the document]
usakhelo: double fails to compile

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+void AppleseedIInteractiveRender::BeginSession()
+{
+    if (m_render_session == nullptr)
+    {
+        RendererSettings renderer_settings = RendererSettings::defaults();
+
+        m_time = GetCOREInterface()->GetTime();
+
+        ViewExp13* vp13 = reinterpret_cast<ViewExp13*>(GetViewExp()->Execute(ViewExp::kEXECUTE_GET_VIEWEXP_13));
```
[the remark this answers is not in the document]
usakhelo: Only this line gets repeated: ViewExp13* vp13 = ... , and it's called on different viewexp-s. Does it worth moving it to a function?

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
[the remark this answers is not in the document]
usakhelo: Got rid of m_currently_rendering. Seems to be working fine..

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.h
```diff
@@ -0,0 +1,96 @@
...
+
+    asf::auto_release_ptr<asr::Project> prepare_project(
+        const RendererSettings&     renderer_settings,
+        const ViewParams&           view_params,
+        const TimeValue             time
+    );
+
+    void update_camera(INode* camera);
+
+    asf::auto_release_ptr<asr::Project> m_project;
```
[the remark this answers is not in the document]
usakhelo: I've made most of them private. update_camera and m_session need to be public

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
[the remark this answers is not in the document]
dictoon: You're right, set_transform() takes a float, I must have been temporarily confused :)

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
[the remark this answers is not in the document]
dictoon: `DWORD` is an integer type, in that case returning 0 is the right thing to do. Basically you should never use `NULL` anymore, it's either 0 or `nullptr`.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.h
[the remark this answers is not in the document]
dictoon: Why does `m_render_session` be public? If it really needs to be, I'd advise adding a method returning a pointer or a (mutable or const) reference, instead of letting the outside access a naked `unique_ptr<>`.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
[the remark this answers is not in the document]
dictoon: I'm surprised that it works at all with 0. Anyway, the good news is that it looks like we can remove completely that second argument (it's a pointer to an integer that represents the number of rendered tiles), because although the integer is incremented, its value is never used.

Regarding this question:

> Not sure if I can pass pointer to a member in a constructor

You totally can, but you have to be very careful about one thing: the _actual_ order of initialization is the order of _declaration_, not the order in which you write the initializations. This is important when you initialize one member with the value of another (that another member must be initialized before you use it).

Example:

    struct C
    {
        int m_a;
        int m_u;

        C()
          : m_u(1)
          , m_a(m_u) // problem: m_a is initialized first since it's declared first, but then m_u is still uninitialized!
        {
        }
}

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -188,16 +185,14 @@ void AppleseedRenderer::GetPlatformInformation(MSTR& info) const
 
 #endif
 
-RefTargetHandle	AppleseedRenderer::Clone(RemapDir &remap)
+RefTargetHandle AppleseedRenderer::Clone(RemapDir& remap)
 {
     AppleseedRenderer* new_rend = static_cast<AppleseedRenderer*>(g_appleseed_renderer_classdesc.Create(false));
-    if (DbgVerify(new_rend != nullptr))
+    if (DbgAssert(new_rend != nullptr))
```
dictoon: You should rely on the return value of `DbgAssert()`.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.h
```diff
@@ -32,12 +59,10 @@ class InteractiveTileCallback
         renderer::IRendererController*  render_controller);
 
     virtual void on_progressive_frame_end(const renderer::Frame* frame) override;
-
-    void update_window();
-    void post_callback(void(*funcPtr)(UINT_PTR), UINT_PTR param);
     static void update_caller(UINT_PTR param_ptr);
 
   private:
+
```
dictoon: Remove newly introduced blank line.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.h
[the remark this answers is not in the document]
usakhelo: made quick fix with function returning a pointer

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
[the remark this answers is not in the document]
usakhelo: I've removed second argument from TileCallback.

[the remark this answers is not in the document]
dictoon: Ah, you're right. Instead of introducing a dummy variable, it would be cleaner to continue passing `nullptr` for this argument (to clearly signify that we don't use it) but to check that the pointer is not null before incrementing the pointed value.

src/appleseed-max-impl/appleseedinteractive/interactivesession.h
```diff
@@ -0,0 +1,70 @@
...
+class InteractiveSession
+{
+  public:
+    InteractiveSession(
+        IIRenderMgr*                iirender_mgr,
+        asr::Project*               project,
+        const RendererSettings&     settings,
+        Bitmap*                     bitmap);
+
+    void render_thread();
```
dictoon: `render_thread()` should be made private.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+    TileCallback::on_progressive_frame_end(frame);
+
+    //wait until ui proc gets handled to ensure class object is valid
+    m_ui_promise = std::promise<void>();
+    if (m_renderer_ctrl->get_status() == asr::IRendererController::ContinueRendering)
+    {
+        std::future<int> ui_future = m_ui_promise.get_future();
+        PostMessage(GetCOREInterface()->GetMAXHWnd(), WM_TRIGGER_CALLBACK, (UINT_PTR)update_caller, (UINT_PTR)this);
+        ui_future.wait();
```
dictoon: I wonder if just doing `m_ui_promise.get_future().wait();` wouldn't be enough?

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    object_ptr->m_ui_promise.set_value();
+}
+
+void InteractiveTileCallback::on_progressive_frame_end(
+    const asr::Frame* frame)
+{
+    TileCallback::on_progressive_frame_end(frame);
+
+    //wait until ui proc gets handled to ensure class object is valid
```
dictoon: Add space after `//`, capitalize first word and end with a period.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.h
```diff
@@ -0,0 +1,70 @@
...
+  : public TileCallback
+{
+  public:
+      InteractiveTileCallback(
+        Bitmap*                         bitmap,
+        IIRenderMgr*                    iimanager,
+        renderer::IRendererController*  render_controller);
+
+    virtual void on_progressive_frame_end(const renderer::Frame* frame) override;
+    static void update_caller(UINT_PTR param_ptr);
```
dictoon: I really don't see why `update_caller()` cannot be made private. What am I missing?

usakhelo: That means you can pass pointer to private methods... Interesting..

dictoon: > That means you can pass pointer to private methods... Interesting..

Yep, private members & methods are just a safeguard for programmers, there is absolutely no runtime checks associated with them (it has zero cost).

In C/C++ you can always get, pass or even craft a pointer to a method and then call it.

PR #54, opened 2017-08-31
Code cleanups

dictoon:


src/appleseed-max-impl/appleseedinteractive/interactivesession.cpp
```diff
@@ -37,22 +37,19 @@
 #include "renderer/api/project.h"
 #include "renderer/api/rendering.h"
 
-// 3ds Max headers.
-#include <interactiverender.h>
-
-// Forward declarations.
-class Bitmap;
+namespace asf = foundation;
+namespace asr = renderer;
```
usakhelo: I could not find any use of asf in this file. But probably should still stay for future use.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
```diff
@@ -37,36 +41,40 @@ namespace
...
-    IIRenderMgr*                    iimanager,
-    asr::IRendererController*       render_controller)
-    : TileCallback(bitmap, &m_rendered_tile_count)
-    , m_bitmap(bitmap)
-    , m_iimanager(iimanager)
-    , m_renderer_ctrl(render_controller)
+    Bitmap*                     bitmap,
+    IIRenderMgr*                iimanager,
+    asr::IRendererController*   render_controller)
+  : TileCallback(bitmap, &m_rendered_tile_count)
```
usakhelo: I should have changed &m_rendered_tile_count to nullptr. Will fix it in next PR.

src/appleseed-max-impl/appleseedinteractive/interactivesession.cpp
dictoon: Oh ok, well we can keep it or remove it, either way it's not a big deal. The main point is to not create namespace aliases in header files.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
dictoon: Cool!

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
```diff
@@ -37,36 +41,40 @@ namespace
...
-    
-    if (object_ptr->m_iimanager->IsRendering())
-        object_ptr->m_iimanager->UpdateDisplay();
-
-    object_ptr->m_ui_promise.set_value();
-}
-
 void InteractiveTileCallback::on_progressive_frame_end(
-    const asr::Frame* frame)
+    const asr::Frame*           frame)
```
usakhelo: Should we insert extra spaces in case of one argument?

PR #57, opened 2017-09-06
Small fixes

dictoon:


dictoon:
Hey @usakhelo, thanks for the review!

You're right, it would probably be better to replace the extension if one exists. I'll try to do that in a subsequent PR. Merging this one.

Thanks!

PR #59, opened 2017-09-07
Force appleseed project file path extension to be .appleseed

dictoon:


src/appleseed-max-impl/utilities.cpp
```diff
@@ -181,3 +181,11 @@ std::string insert_texture_and_instance(
 
     return texture_instance_name;
 }
+
+WStr replace_extension(const WStr& file_path, const WStr& new_ext)
+{
+    const int i = file_path.first(L'.');
```
usakhelo: Is this replaces everything after first period?
I would suggest to look for the last dot in filename and replace everything after that or append extension if there's no period. In case of the only dot in file it will be the same, but if someone wants to have periods in filename it may help.

dictoon: Oh, you're totally right, I meant `last()` and not `first()`...

PR #61, opened 2017-09-14
Modernize string handling

dictoon:


dictoon:
Yeah, `TCHAR` and `_T()` were invented to simultaneously support Unicode and non-Unicode versions of Windows. There is no point anymore to support non-Unicode versions (that would be Windows 98 and earlier).

You can read more about this topic on StackOverflow:
https://stackoverflow.com/questions/234365/is-tchar-still-relevant
https://stackoverflow.com/questions/6315566/should-i-eliminate-tchar-from-windows-code

dictoon:
Thanks for the review!

PR #63, opened 2017-09-20
Fix two bugs

### LINKED ISSUE #62: Maps slots shown incorrectly on Glass and SSS materials in Slate material editor mode
This affects map slot indexing. For example in Glass material if map is shown as connected to Bump Map slot it actually connected to Volume Color slot.
image

dictoon:
This fixes #62.


dictoon:
Yep, compiled and tested with Max 2015.

Thanks for the review!

PR #64, opened 2017-09-21
Support 3ds max procedural maps via custom source classes and some UI enhancement

usakhelo:
Hi @dictoon,
I think I'm ready for new PR. All the materials have been modified to support max procedural maps using custom sources.
I've added new checkbox in the System rollout to switch between normal and non-compatible mode. In the new mode project files are not saved.
Also unneeded dialog tabs (Raytrace, Advanced lighting) are not added anymore to the render dialog.
Maybe ShadingContext and custom Source and Texture classes should be moved to separate file from utilities.cpp.


src/appleseed-max-impl/appleseedglassmtl/appleseedglassmtl.h
```diff
@@ -132,9 +132,17 @@ class AppleseedGlassMtl
     virtual bool can_emit_light() const override;
     virtual foundation::auto_release_ptr<renderer::Material> create_material(
         renderer::Assembly& assembly,
-        const char*         name) override;
+        const char*         name,
+        bool                use_max_source) override;
```
dictoon: Maybe `use_max_procedural_maps`?

Also: make const.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -389,7 +389,8 @@ namespace
         asr::Assembly&          assembly,
         const std::string&      instance_name,
         Mtl*                    mtl,
-        MaterialMap&            material_map)
+        MaterialMap&            material_map,
+        const bool              use_max_proc_maps)
```
dictoon: That's definitely a better name! As I mentioned in another comment, I'd go even further (and use the same name everywhere): `use_max_procedural_maps`

src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
```diff
@@ -763,18 +776,135 @@ asf::auto_release_ptr<asr::Material> AppleseedDisneyMtl::create_material(asr::As
...
+        else bsdf_params.insert("roughness", m_roughness / 100.0f);
+
+        // Anisotropic.
+        if (is_supported_texture(m_anisotropy_texmap))
+            bsdf_params.insert("anisotropic", insert_max_texture_and_instance(assembly, m_anisotropy_texmap));
+        else bsdf_params.insert("anisotropic", m_anisotropy);
+
+        const auto bsdf_name = std::string(name) + "_bsdf";
+        assembly.bsdfs().insert(
+            asr::DisneyBRDFFactory::static_create(bsdf_name.c_str(), bsdf_params));
```
dictoon: Awesome, that was exactly the right thing to do!

src/appleseed-max-impl/utilities.cpp
```diff
@@ -47,13 +48,387 @@
...
+            alpha.set(evaluate_float(uv));
+        }
+
+        virtual void evaluate(
+            asr::TextureCache&          texture_cache,
+            const asf::Vector2f&        uv,
+            asf::Color3f&               linear_rgb,
+            asr::Alpha&                 alpha) const override
+        {
+            const asf::Color4f color = evaluate_color(uv);
```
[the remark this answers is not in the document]
usakhelo: Great point, thanks.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -319,6 +318,16 @@ void AppleseedInteractiveRender::BeginSession()
     else
         get_view_params_from_viewport(view_params, *GetViewExp(), m_time);
 
+    RendererSettings renderer_settings = RendererSettings::defaults();
+    
+    Renderer* curr_renderer = GetCOREInterface()->GetCurrentRenderer(false);
+    if (curr_renderer && curr_renderer->ClassID() == AppleseedRenderer::get_class_id())
```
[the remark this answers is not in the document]
usakhelo: Probably cannot happen. Should I remove whole if statement or should still check `if (curr_renderer)`?

usakhelo:
@dictoon, I'm still working on changes, will commit final version later.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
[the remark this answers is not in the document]
dictoon: If it cannot happen for sure, you should remove the entire statement.

In fact, having a closer look at it, the entire statement is a no-op since `AppleseedRenderer::get_renderer_settings()` returns a _copy_ of the renderer settings, which you then modify, which then get lost. So the whole thing is doing nothing :)

src/appleseed-max-impl/utilities.cpp
```diff
@@ -277,10 +290,8 @@ namespace
             asr::Spectrum&              spectrum) const override
         {
             DbgAssert(spectrum.size() == 3);
-            const asf::Color4f color = evaluate_color(uv);
-            spectrum[0] = color[0];
-            spectrum[1] = color[1];
-            spectrum[2] = color[2];
+            asf::Color3f color;
```
dictoon: You can remove `color` here.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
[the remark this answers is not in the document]
usakhelo: Hi @dictoon, I only modify one setting after I get renderer_settings from main renderer, all others are kept aren't they? I certainly see them carried from main renderer when I change settings in the dialog.
Function declared like this
`const RendererSettings& AppleseedRenderer::get_renderer_settings()` It returns a reference I think.. But the assignment makes a copy, is that what you mean?
Should I do 
`RendererSettings& renderer_settings = appleseed_renderer->get_renderer_settings();` instead?

src/appleseed-max-impl/appleseedglassmtl/appleseedglassmtl.h
dictoon: Comment still holds :)

src/appleseed-max-impl/appleseedglassmtl/appleseedglassmtl.cpp
```diff
@@ -631,16 +631,8 @@ asf::auto_release_ptr<asr::Material> AppleseedGlassMtl::create_material(
...
-        return create_max_material(assembly, name);
-    else
-        return create_universal_material(assembly, name);
-}
-
-asf::auto_release_ptr<asr::Material> AppleseedGlassMtl::create_universal_material(asr::Assembly& assembly, const char* name)
-{
     asr::ParamArray material_params;
-
+    std::string texture_name;
```
dictoon: Put back a blank line before the comment block below.

src/appleseed-max-impl/appleseedsssmtl/appleseedsssmtl.cpp
```diff
@@ -742,135 +739,22 @@ asf::auto_release_ptr<asr::Material> AppleseedSSSMtl::create_universal_material(
     // Material.
     //
 
-    if (is_bitmap_texture(m_bump_texmap))
+    texture_name = insert_texture_and_instance(
+        assembly,
+        m_bump_texmap,
+        use_max_procedural_maps,
+        asr::ParamArray()
+        .insert("color_space", "linear_rgb"));
```
dictoon: Indent this line.

src/appleseed-max-impl/utilities.cpp
```diff
@@ -567,6 +567,7 @@ std::string insert_texture_and_instance(
 {
     std::string texture_instance_name;
     std::string texture_name;
+
     if (use_max_source)
```
dictoon: Let's use the same name everywhere for this argument: `use_max_procedural_maps`

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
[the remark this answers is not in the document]
dictoon: The way you ended up doing it is correct:

- `AppleseedRenderer::get_renderer_settings()` returns a constant reference to the settings, i.e. it gives _read only_ access to the settings (which is the right thing to do, we don't want users of this class to be able to modify the renderer settings).

- When you do

      RendererSettings renderer_settings = AppleseedRenderer::get_renderer_settings();

  the settings are indeed copied.

- This wouldn't compile at all:

      RendererSettings& renderer_settings = AppleseedRenderer::get_renderer_settings();

  You can't convert a constant reference to a mutable (non-constant) one.

src/appleseed-max-impl/appleseeddisneymtl/appleseeddisneymtl.cpp
```diff
@@ -792,7 +792,9 @@ asf::auto_release_ptr<asr::Material> AppleseedDisneyMtl::create_max_material(
     asr::Assembly&  assembly,
     const char*     name)
 {
-     asr::ParamArray material_params;
+    asr::ParamArray material_params;
+    std::string texture_name;
+    bool use_max_procedural_maps = true;
```
usakhelo: Hi @dictoon, this variable is used in place of bool value `true` in all `insert_texture_and_instance()` calls. Is it ok or i should be using `true` instead?

dictoon:
Rebased and merged, thank you!!

PR #65, opened 2017-09-24
Refactorings and cleanups

dictoon:


src/appleseed-max-impl/utilities.cpp
```diff
@@ -57,33 +57,233 @@
...
+{
+    if (map == nullptr)
+        return false;
+
+    auto part_a = map->ClassID().PartA();
+    auto part_b = map->ClassID().PartB();
+
+    switch (part_a)
+    {
+      case 0x64035FB9:              // Tiles.
```
usakhelo: Those comments should be without the periods probably?

dictoon: Good point!

PR #76, opened 2017-11-06
Refactor setup_environment_map function

usakhelo:


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1027,49 +1027,43 @@ namespace
...
 
                 if (sun_dir.y > 0.0f)
                     sun_phi = asf::TwoPi<float>() - sun_phi;
 
-                sun_theta = asf::rad_to_deg(sun_theta);
-                sun_phi = asf::rad_to_deg(sun_phi);
-            }
-            else
-            {
-                get_paramblock_value_by_name(param_block, L"sun_theta", time, sun_theta, FOREVER);
```
dictoon: So this code is no longer needed?

usakhelo: Yes, `Update()` does that now.
Here's what docs say about material's `Update()` function:
>....This method is called before rendering begins to allow the plug-in to evaluate anything prior to the render so it can store this information. .... It is generally called once at the start of each frame, or during interactive playback when the time changes.
So calling update before the rendering seems correct.

usakhelo:
Update() seems to work correctly without calling it explicitly. It's get called each time parameter changes and so on.. Which is what you expected it to do.

dictoon:
Looks good, thanks!

PR #84, opened 2017-11-27
Create and connect uv_transform shader when adding textures

usakhelo:
This PR adds [uv_transform shader](https://github.com/appleseedhq/appleseed/pull/1737) support to `oslutils.cpp`
Tested on 2016 and 2017. Works on both color and float textures.

### DIFF
--- a/src/appleseed-max-impl/oslutils.cpp
+++ b/src/appleseed-max-impl/oslutils.cpp
@@ -41,14 +41,116 @@
 
 // 3ds Max Headers.
 #include <bitmap.h>
+#include <imtl.h>
+#include <maxapi.h>
 #include <stdmat.h>
 
 namespace asf = foundation;
 namespace asr = renderer;
 
+namespace
+{
+    asr::ParamArray get_uv_params(Texmap* texmap)
+    {
+        asr::ParamArray uv_params;
+
+        if (!is_bitmap_texture(texmap))
+            return uv_params;
+
+        UVGen* uv_gen = texmap->GetTheUVGen();
+        if (!uv_gen || !uv_gen->IsStdUVGen())
+            return uv_params;
+
+        StdUVGen* std_uv = static_cast<StdUVGen*>(uv_gen);
+        auto time = GetCOREInterface()->GetTime();
+
+        // Assume that all the textures are in texture mode texmap->MapSlotType() == MAPSLOT_TEXTURE.
+        // Assume that static_cast<StdUVGen*>(uv_gen)->GetCoordMapping() == UVMAP_EXPLICIT.
+        float u_tiling = std_uv->GetUScl(time);
+        float v_tiling = std_uv->GetVScl(time);
+        float u_offset = std_uv->GetUOffs(time);
+        float v_offset = std_uv->GetVOffs(time);
+        float w_rotation = std_uv->GetWAng(time);
+        int tiling = std_uv->GetTextureTiling();
+
+        if (tiling & U_WRAP)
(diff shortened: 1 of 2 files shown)


usakhelo:
Just released new function should be local utility function. Will update later.

usakhelo:
Hi,
I've updated the commit. There was a bug and now it's fixed and ready for review.

src/appleseed-max-impl/oslutils.cpp
```diff
@@ -41,16 +41,118 @@
...
+            return uv_params;
+
+        UVGen* uv_gen = texmap->GetTheUVGen();
+        if (!uv_gen || !uv_gen->IsStdUVGen())
+            return uv_params;
+
+        StdUVGen* std_uv = static_cast<StdUVGen*>(uv_gen);
+        auto time = GetCOREInterface()->GetTime();
+
+        // Assume that all the textures are in texture mode texmap->MapSlotType() == MAPSLOT_TEXTURE.
```
dictoon: Even better than comments: assertions. Something like

    DbgAssert(texmap->MapSlotType() == MAPSLOT_TEXTURE);
    DbgAssert(static_cast<StdUVGen*>(uv_gen)->GetCoordMapping() == UVMAP_EXPLICIT);

src/appleseed-max-impl/oslutils.cpp
```diff
@@ -41,16 +41,118 @@
...
+
+        uv_params.insert("in_offsetU", fmt_osl_expr(u_offset));
+        uv_params.insert("in_offsetV", fmt_osl_expr(v_offset));
+
+        uv_params.insert("in_tilingU", fmt_osl_expr(u_tiling));
+        uv_params.insert("in_tilingV", fmt_osl_expr(v_tiling));
+
+        uv_params.insert("in_rotateW", fmt_osl_expr(w_rotation));
+
+        // Access BMTex paramaters which are only accessible in a weird way.
```
dictoon: *parameters*

dictoon: What's the weird way of accessing them?

src/appleseed-max-impl/oslutils.cpp
```diff
@@ -41,16 +41,118 @@
...
+            bmtex_jitter,
+            bmtex_usejitter,
+            bmtex_apply,
+            bmtex_crop_place
+        };
+
+        auto pblock = texmap->GetParamBlock(bmtex_params);
+        if (pblock)
+        {
+            int use_clip = pblock->GetInt(bmtex_apply, time);
```
dictoon: Make all local variables const.

src/appleseed-max-impl/oslutils.cpp
usakhelo: It looked weird to me to not have get/set methods. Changed to something neutral, but I can remove it completely if you think it's not needed.

src/appleseed-max-impl/oslutils.cpp
usakhelo: Great advice thanks. And in release build those asserts are just ignored? That's what I wanted to happen.

dictoon: Yes, assertions are completely ignored in optimized builds. Which poses a risk, by the way, for instance this wouldn't be a proper way to check that `write_to_file()` worked, because the entire call would be ignored in optimized builds:

    DbgAssert(write_to_file("file.ext"));

Something to keep in mind :)

PR #98, opened 2017-12-06
Remove mix color swatches, further fix blend with duplicate sub-materials

usakhelo:
Color swatches are removed. Number amount is texture multiplier when texture presented and material mixing amount without texture.
Also fix blend material in cases when sub-materials have duplicate names. Previous fix failed when sub-material is another blend material with the same name.


src/appleseed-max-impl/appleseedblendmtl/appleseedblendmtl.rc
```diff
@@ -56,71 +56,62 @@ FONT 8, "MS Sans Serif", 0, 0, 0x1
...
     CONTROL         "Layer Material 1",IDC_MTLBTN_LAYER_1,"CustButton",WS_TABSTOP,18,29,79,10
     CONTROL         "Texture Amount Edit 1",IDC_EDIT_AMOUNT_1,"CustEdit",WS_TABSTOP,179,29,25,10
     CONTROL         "Texture Amount Spinner 1",IDC_SPIN_AMOUNT_1,
                     "SpinnerControl",0x0,205,29,7,10
-    CONTROL         "Mix Color 1",IDC_MIX_AMOUNT_1,"ColorSwatch",WS_TABSTOP,100,29,10,10
     LTEXT           "1:",IDC_LABEL_COAT_1,9,30,8,8
     LTEXT           "Layers:",IDC_LABEL_BASE_MTL2,18,20,24,8
     LTEXT           "Blend Amount:",IDC_LABEL_BASE_MTL3,100,20,52,8
-    CONTROL         "Mask Texture 1",IDC_TEXBTN_MASK_2,"CustButton",WS_TABSTOP,112,42,65,10
+    CONTROL         "Mask Texture 1",IDC_TEXBTN_MASK_2,"CustButton",WS_TABSTOP,100,42,76,10
```
dictoon: Fix control names.

src/appleseed-max-impl/oslutils.cpp
```diff
@@ -353,7 +353,7 @@ void connect_sub_mtl(
         return;
 
     std::string layer_name =
-        make_unique_name(assembly.materials(), wide_to_utf8(mat->GetName()) + "_sub_mat");
+        make_unique_name(assembly.materials(), asf::format("{0}_{1}_{2}", shader_name, mat->GetName(), "sub_mat"));
```
dictoon: "sub_mat" can be part of the format string.

dictoon:
Thanks man!

src/appleseed-max-impl/oslutils.cpp
usakhelo: such a stupid mistake really :)

usakhelo:
Found one more bug in blend material - max crushes when empty blend material is used as a base sub-material. Now it's fixed. And resource file is cleaned up.
Thanks for the review.

src/appleseed-max-impl/oslutils.cpp
dictoon: It's not even really a mistake :)

dictoon:
Thanks!!

PR #100, opened 2017-12-08
Light and sss osl shaders

usakhelo:
Also simplifies blend material compatibility check - blend material discards any material without osl_surface, so checking the type is redundant, I think.
Improves lighting materials check in the scene by checking blend's submaterials.
Makes blend material look in viewport similar to it's base material.

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -924,33 +924,35 @@ namespace
         }
     }
 
+    bool is_light_emitting_material(Mtl* mtl)
+    {
+        IAppleseedMtl* appleseed_mtl =
+            static_cast<IAppleseedMtl*>(mtl->GetInterface(IAppleseedMtl::interface_id()));
+        if (appleseed_mtl != nullptr && appleseed_mtl->can_emit_light())
+            return true;
+        else
+        {
+            const int submtl_count = mtl->NumSubMtls();
+            for (int i = 0; i < submtl_count; ++i)
+            {
+                Mtl* sub_mtl = mtl->GetSubMtl(i);
+                if (sub_mtl != nullptr)
+                {
+                    if (is_light_emitting_material(sub_mtl))
+                        return true;
+                }
+            }
+        }
+        return false;
+    }
+
     bool has_light_emitting_materials(const MaterialMap& material_map)
     {
         for (const auto& entry : material_map)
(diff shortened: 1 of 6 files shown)


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -924,15 +924,37 @@ namespace
...
     bool has_light_emitting_materials(const MaterialMap& material_map)
     {
         for (const auto& entry : material_map)
         {
             Mtl* mtl = entry.first;
-            IAppleseedMtl* appleseed_mtl =
-                static_cast<IAppleseedMtl*>(mtl->GetInterface(IAppleseedMtl::interface_id()));
-            if (appleseed_mtl->can_emit_light())
-                return true;
+            return is_light_emitting_material(mtl);
```
dictoon: That doesn't look right. It should most likely be:

    if (is_light_emitting_material(mtl))
        return true;

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -924,15 +924,37 @@ namespace
...
+    bool is_light_emitting_material(Mtl* mtl)
+    {
+        IAppleseedMtl* appleseed_mtl =
+            static_cast<IAppleseedMtl*>(mtl->GetInterface(IAppleseedMtl::interface_id()));
+        if (appleseed_mtl != nullptr && appleseed_mtl->can_emit_light())
+            return true;
+        else
+        {
+            const int submtl_count = mtl->NumSubMtls();
+            if (submtl_count > 0)
```
dictoon: This check is redundant.

usakhelo:
@dictoon what about these changes, are they good to merge?

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -924,14 +924,34 @@ namespace
         }
     }
 
+    bool is_light_emitting_material(Mtl* mtl)
+    {
+        IAppleseedMtl* appleseed_mtl =
+            static_cast<IAppleseedMtl*>(mtl->GetInterface(IAppleseedMtl::interface_id()));
+        if (appleseed_mtl != nullptr && appleseed_mtl->can_emit_light())
+            return true;
+        else
```
dictoon: Note for later: this `else` can be removed.

PR #105, opened 2017-12-19
Fix bump in "use procedural maps" mode

usakhelo:
Hi @dictoon,
I think this fixes bump and normal problem when renderer is in "use max procedural maps" mode. At least bump and normal are showing in render, both with bitmaps and procedural textures.
I remember we use some dummy values for procedural textures, and then this offset was divided by that value. But now for some reason division is not required.


dictoon:
Good catch. I don't remember either why the offset was divided by 512, but clearly it shouldn't.

PR #113, opened 2017-12-26
Add dialog window for rendering log

usakhelo:
Hi,
This pr adds dialog window to print rendering log messages. Three ui elements added to the System rollout of appleseed settings tab:
 - ComboBox with three choices of messages types that will cause dialog to open automatically
 - Button to open dialog manually before render or open dialog with messages from previous rendering session
 - Checkbox to turn on/off logging of rendering of material samples

Combobox and checkbox state is saved into ini file to standard max config folder and loaded on each max startup - %localappdata%\Autodesk\3dsMax\2016 - 64bit\ENU\en-US\plugcfg\appleseed\appleseed.ini (language code and max version varies).

Probably worth noting: log window is another appleseed log target that gets attached and detached on rendering start and finish, so no plugin initialization warnings and errors will show up in the window.

Thanks for review,
Sergo.


dictoon:
Frickin' awesome Sergo!!

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.rc
```diff
@@ -119,16 +119,29 @@ BEGIN
...
     CONTROL         "CPU Cores",IDC_SPINNER_RENDERINGTHREADS,"SpinnerControl",WS_TABSTOP,71,5,6,10
-    CONTROL         "Low Priority Mode",IDC_CHECK_LOW_PRIORITY_MODE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,0,36,73,10
+    CONTROL         "Low Priority Mode",IDC_CHECK_LOW_PRIORITY_MODE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,0,58,73,10
     CONTROL         "Use Max Procedural Maps (non-compatible mode)",IDC_CHECK_USE_MAX_PROCEDURAL_MAPS,
-                    "Button",BS_AUTOCHECKBOX | WS_TABSTOP,0,24,197,10
+                    "Button",BS_AUTOCHECKBOX | WS_TABSTOP,0,48,197,10
+    COMBOBOX        IDC_COMBO_LOG,39,20,65,30,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    LTEXT           "Open Log:",IDC_STATIC_LOG,0,23,34,8
+    CONTROL         "Open Log",IDC_BUTTON_LOG,"CustButton",WS_TABSTOP,110,20,42,12
+    CONTROL         "Log material slots rendering ",IDC_CHECK_LOG_MATERIAL_EDITOR,
```
dictoon: Remove space at the end of string.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -712,8 +716,14 @@ namespace
             m_spinner_renderingthreads->SetResetValue(RendererSettings::defaults().m_rendering_threads);
             m_spinner_renderingthreads->SetValue(m_settings.m_rendering_threads, FALSE);
 
+            const wchar_t* log_combo_items[] = { L"Always", L"Never", L"On Error" };
```
dictoon: Mark as `static`. Also, it's an array of constants so `LogComboItems`, according to our conventions.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -712,8 +716,14 @@ namespace
...
+            for (int i = 0; i < 3; i++)
```
dictoon: Prefer `size_t` unless another type is required.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -735,6 +745,27 @@ namespace
...
+                    m_renderer->show_last_session_log();
+                    return TRUE;
+                  
+                  case IDC_COMBO_LOG:
+                    if (HIWORD(wparam) == CBN_SELCHANGE)
+                    {
+                        LRESULT sel_mode = SendDlgItemMessage(hwnd, IDC_COMBO_LOG, CB_GETCURSEL, 0, 0);
+                        if (sel_mode != CB_ERR)
+                        {
+                            m_settings.m_log_open_mode = static_cast<DialogLogMode>((int)sel_mode);
```
dictoon: Is the cast to `int` required here?

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.h
```diff
@@ -39,6 +39,7 @@
 
 // Forward declarations.
 class RendererSettings;
+class AppleseedRenderer;
```
dictoon: Order alphabetically.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include <Richedit.h>
+
+namespace asf = foundation;
+namespace asr = renderer;
+
+namespace
+{
+    boost::mutex                g_message_queue_mutex;
+    std::vector<MessagePair>    g_message_queue;
+    HWND                        g_log_dialog = 0;
```
dictoon: I believe we should use `nullptr` here since ultimately `HWND` is a pointer (to `void`)

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+    static INT_PTR CALLBACK dialog_proc(HWND hwnd, UINT msg, WPARAM wparam, LPARAM lparam)
+    {
+        switch (msg)
+        {
+          case WM_INITDIALOG:
+            break;
+          case WM_CLOSE:
+          case WM_DESTROY:
+            if (g_log_dialog != 0)
```
dictoon: `nullptr` here too and in other places as well.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
```diff
@@ -0,0 +1,259 @@
...
+            {
+                GetCOREInterface14()->UnRegisterModelessRenderWindow(g_log_dialog);
+                g_log_dialog = 0;
+                DestroyWindow(hwnd);
+            }
+            break;
+          
+          case WM_SIZE:
+          {
```
dictoon: Ident `{` like any other code under `case` (2 spaces). Then the contents of the `{}` block is indented like any other code (4 spaces).

    case WM_SIZE:
      {
          HWND edit_box = GetDlgItem(hwnd, IDC_EDIT_LOG);
          ...
      }
      break;

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
```diff
@@ -0,0 +1,259 @@
...
+        const int text_length = GetWindowTextLength(edit_box);
+        SendMessage(edit_box, EM_SETSEL, text_length, text_length);
+
+        CHARFORMAT char_format;
+        SendMessage(edit_box, EM_GETCHARFORMAT, SCF_SELECTION, reinterpret_cast<LPARAM>(&char_format));
+        char_format.cbSize = sizeof(char_format);
+        char_format.dwMask = CFM_COLOR | CFM_FACE;
+        char_format.dwEffects = 0;
+        char_format.crTextColor = color;
+        wcscpy(reinterpret_cast<WCHAR*>(char_format.szFaceName), L"Consolas");
```
dictoon: Is the `reinterpret_cast<>` required here? `CHARFORMAT.szFaceName` appears to be of type `TCHAR` of which the [documentation](https://msdn.microsoft.com/en-us/library/office/cc842072.aspx) says:

> For Unicode platforms, TCHAR is defined as synonymous with the WCHAR type.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
```diff
@@ -0,0 +1,259 @@
...
+                  default:
+                    message_color = RGB(10, 10, 10);
+                    break;
+                }
+                append_text(GetDlgItem(g_log_dialog, IDC_EDIT_LOG), utf8_to_wide(line + "\n").c_str(), message_color);
+            }
+        }
+    }
+
+    void push_message(MessagePair message)
```
dictoon: Take a const ref.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
```diff
@@ -0,0 +1,259 @@
...
+                g_module,
+                MAKEINTRESOURCE(IDD_DIALOG_LOG),
+                GetCOREInterface()->GetMAXHWnd(),
+                dialog_proc,
+                NULL);
+
+            GetCOREInterface14()->RegisterModelessRenderWindow(g_log_dialog);
+        }
+
+        if (g_log_dialog != NULL)
```
dictoon: Use `nullptr`.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+}
+
+void DialogLogTarget::show_last_session_messages()
+{
+    if (g_log_dialog)
+        return;
+
+    boost::mutex::scoped_lock lock(g_message_queue_mutex);
```
dictoon: Wrap this block in curly braces such that the mutex gets released before the call to `PostMessage()`.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.h
```diff
@@ -0,0 +1,71 @@
...
+#include "foundation/utility/log.h"
+
+// Standard headers.
+#include <cstddef>
+#include <string>
+#include <vector>
+
+typedef std::vector<std::string> StringVec;
+typedef foundation::LogMessage::Category MessageType;
+typedef std::pair<MessageType, StringVec> MessagePair;
```
dictoon: Include <utility> for `std::pair<>`.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.h
```diff
@@ -0,0 +1,71 @@
...
+
+enum class DialogLogMode
+{
+    Always,
```
dictoon: Assign explicit values and add the usual comment that says that changing these values will break compatibility (since they are saved as is to Max files).

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.h
```diff
@@ -0,0 +1,71 @@
...
+    Never,
+    Errors
+};
+
+class DialogLogTarget
+    : public foundation::ILogTarget
```
dictoon: Half-indent.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.h
```diff
@@ -0,0 +1,71 @@
...
+{
+public:
```
dictoon: Half-indent.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.h
```diff
@@ -0,0 +1,71 @@
...
+        const MessageType           category,
+        const char*                 file,
+        const size_t                line,
+        const char*                 header,
+        const char*                 message) override;
+
+    void print_to_dialog();
+    void show_last_session_messages();
+
+private:
```
dictoon: Half-indent.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.h
```diff
@@ -0,0 +1,71 @@
...
+    Always,
+    Never,
+    Errors
+};
+
+class DialogLogTarget
+    : public foundation::ILogTarget
+{
+public:
+    DialogLogTarget(DialogLogMode   open_mode);
```
dictoon: Make constructor explicit. Make argument const.

src/appleseed-max-impl/utilities.h
```diff
@@ -329,3 +348,52 @@ std::string make_unique_name(
...
+    filename += GetCOREInterface()->GetDir(APP_PLUGCFG_DIR);
+    filename += L"\\appleseed\\appleseed.ini";
+
+    wchar_t buf[100];
+    int count = GetPrivateProfileString(
+        category,
+        key_name,
+        utf8_to_wide(foundation::to_string(default_value)).c_str(),
+        buf,
+        100,
```
dictoon: Use a constant for the max number of characters in `buf`.

Are you sure you need to pass 100 and not 99?

Is the buffer large enough?

src/appleseed-max-impl/utilities.h
```diff
@@ -329,3 +348,52 @@ std::string make_unique_name(
...
+        filename);
+
+    std::wstring result;
+    result.assign(buf);
```
dictoon: const  std::wstring result(buf);

should work.

src/appleseed-max-impl/utilities.h
```diff
@@ -329,3 +348,52 @@ std::string make_unique_name(
...
+template <typename T>
+const T load_system_setting(const wchar_t* key_name, const T& default_value)
+{
+    return load_ini_setting(L"System", key_name, default_value);
+}
+
+template <typename T>
+void save_ini_setting(const T& value, const wchar_t* category, const wchar_t* key_name)
+{
+    MaxSDK::Util::Path filename(GetCOREInterface()->GetDir(APP_PLUGCFG_DIR));
```
dictoon: `filename` -> `filepath`.

src/appleseed-max-impl/utilities.h
```diff
@@ -190,6 +192,23 @@ std::string insert_procedural_texture_and_instance(
     renderer::ParamArray    texture_instance_params = renderer::ParamArray());
 
 
+//
+// Plugcfg ini file access functions.
+//
+
+template <typename T>
+void save_ini_setting(const T& value, const wchar_t* category, const wchar_t* key_name);
```
dictoon: Let's always use the same order: category, key name, value.

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -67,6 +67,10 @@ namespace
             m_rendering_threads = 0;    // 0 = as many as there are logical cores
             m_low_priority_mode = true;
             m_use_max_procedural_maps = false;
+
+            const int log_mode = load_system_setting(L"LogOpenMode", static_cast<int>(DialogLogMode::Always));
```
dictoon: What happens if you remove the static cast and convert `log_mode` to `DialogLogMode`? What about `load_system_setting<DialogLogMode>()`?

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.h
```diff
@@ -0,0 +1,71 @@
...
+// Standard headers.
+#include <cstddef>
+#include <string>
+#include <vector>
+
+typedef std::vector<std::string> StringVec;
+typedef foundation::LogMessage::Category MessageType;
+typedef std::pair<MessageType, StringVec> MessagePair;
+
+enum class DialogLogMode
```
dictoon: Shouldn't it be `LogDialogMode` ("mode of the dialog of logs")?

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -319,10 +319,13 @@ void AppleseedInteractiveRender::BeginSession()
         get_view_params_from_viewport(view_params, *GetViewExp(), m_time);
 
     Renderer* curr_renderer = GetCOREInterface()->GetCurrentRenderer(false);
+    // This is not production appleseed renderer, this is completetly independent renderer object.
```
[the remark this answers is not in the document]
usakhelo: Just wanted to have a reminder here because I've spent some time to figure out which object is accessed here. I can remove it. Or add "Reminder:"

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
usakhelo: Thanks! This part was very confusing.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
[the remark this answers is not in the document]
dictoon: Sure, but what does the comment mean? What do you mean by "production appleseed renderer"?

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
usakhelo: Cool! I could avoid extra assignment.

src/appleseed-max-impl/utilities.h
usakhelo: It's enough for now I think, we only store "false/true" and int 0-2.

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
usakhelo: ah, no.. looks like `T from_string(const std::string& s)` can't accept DialogLogMode.
Should I make `m_settings.m_log_open_mode` an int?

dictoon: I was suspecting that it wouldn't work. Let's keep it as you did (without the superfluous casts, if any).

src/appleseed-max-impl/utilities.h
usakhelo: Do you mean I should consider \0 at the end of string? I'm not sure. `GetPrivateProfileString`'s parameter is nSize. It says string is truncated if buffer is too small. It probably considers \0-s in the end?

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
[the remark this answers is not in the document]
usakhelo: I thought it's confusing because I started with non-interactive renderer (production) and then switched to interactive. Then I saw that line and I thought it tries to access production renderer to take it's settings. But in reality it's accessing interactive renderer which is also `AppleseedRender` object. That object later produces `AppleseedInteractiveRender`.
It is all very confusing when you forgot about it. Probably it's clear enough if I think carefully. I can remove that comment, it doesn't make things clearer. I'll add something better next time it confuses me.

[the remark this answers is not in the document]
usakhelo: So line `RendererSettings renderer_settings = appleseed_renderer->get_renderer_settings();` basically accesses default settings. But it looks like it tries to copy settings from somewhere, right?

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.h
```diff
@@ -35,37 +35,40 @@
...
     void print_to_dialog();
     void show_last_session_messages();
 
-private:
-    std::vector<MessagePair>        m_session_messages;
-    DialogLogMode                   m_log_mode;
+  private:
+    std::vector<MessagePair>    m_session_messages;
+    LogDialogMode               m_log_mode;
 };
```
usakhelo: Should I line up all the member along constructor argument?

dictoon: Up to you, whatever looks the nicest :)

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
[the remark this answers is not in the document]
dictoon: Regarding production vs interactive renderer: I understand now. It makes a lot of sense to explain this in a comment.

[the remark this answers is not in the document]
dictoon: Regarding renderer settings: I'm not sure to understand the question, but 

    RendererSettings renderer_settings = appleseed_renderer->get_renderer_settings();

clearly _copies_ the renderer settings into the local `renderer_settings` variable.

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
```diff
@@ -183,7 +183,7 @@ namespace
     const UINT WM_TRIGGER_CALLBACK = WM_USER + 4764;
 }
 
-DialogLogTarget::DialogLogTarget(DialogLogMode open_mode)
+DialogLogTarget::DialogLogTarget(LogDialogMode open_mode)
```
dictoon: Add `const` here too.

src/appleseed-max-impl/utilities.h
```diff
@@ -377,23 +377,23 @@ const T load_system_setting(const wchar_t* key_name, const T& default_value)
 }
 
 template <typename T>
-void save_ini_setting(const T& value, const wchar_t* category, const wchar_t* key_name)
+void save_ini_setting(const wchar_t* category, const wchar_t* key_name, const T& value)
```
dictoon: You need to update the forward declarations of these functions too.

usakhelo: Terrible mistake!

PR #114, opened 2017-12-27
Fix DOF in activeshader rendering, update camera parameters

### LINKED ISSUE #69: DOF not working in ActiveShade rendering

usakhelo:
Hi,
This should fix #69. 
Also fixes updating camera parameters during interactive rendering, such as FOV, aperture, blades and so on. Camera type cannot be changed so switching Enable Depth of Field checkbox won't change anything during the rendering.

Thanks,
Sergo.

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -1292,6 +1292,33 @@ namespace
     }
 }
 
+void set_camera_dof_params(
+    asr::ParamArray&            params,
+    MaxSDK::IPhysicalCamera*    camera_node,
+    Bitmap*                     bitmap,
+    const TimeValue             time)
+{
+    // Film dimensions.
+    const float aspect = static_cast<float>(bitmap->Height()) / bitmap->Width();
+    const float film_width = camera_node->GetFilmWidth(time, FOREVER);
+    const float film_height = film_width * aspect;
+    params.insert("film_dimensions", asf::Vector2f(film_width, film_height));
+
+    // DOF settings.
+    params.insert("f_stop", camera_node->GetLensApertureFNumber(time, FOREVER));
+    params.insert("focal_distance", camera_node->GetFocusDistance(time, FOREVER));
+    switch (camera_node->GetBokehShape(time, FOREVER))
+    {
+      case MaxSDK::IPhysicalCamera::BokehShape::Circular:
+        params.insert("diaphragm_blades", 0);
+        break;
+      case MaxSDK::IPhysicalCamera::BokehShape::Bladed:
+        params.insert("diaphragm_blades", camera_node->GetBokehNumberOfBlades(time, FOREVER));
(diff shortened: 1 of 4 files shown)


usakhelo:
Hi @dictoon,
Question on updating camera type - I've checked appleseed studio code and it seems that when I change camera type there, it removes old camera, creates new one and adds it to the `scene->cameras()` container. I've tried to do the same in max, but it always crashes. I see that studio schedules those changes to happen between `frame_end` and `render_begin`. Since there's no such things in max plugin yet, I've tried to pause the rendering, clear cameras(), add new camera, and then reinitialize and resume the rendering. It's still crashing. Do you think it is timing problem or some other things should also happen to update the camera? I was using `build_camera()` function to create new camera. And I think I can add simple task scheduling to render_controller and session_manager to make it work in the same manner as in studio. I was just wandering if I'm missing something more important.
Sorry for long question.
Thanks,
Sergo.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1292,6 +1292,33 @@ namespace
...
+    const float film_width = camera_node->GetFilmWidth(time, FOREVER);
+    const float film_height = film_width * aspect;
+    params.insert("film_dimensions", asf::Vector2f(film_width, film_height));
+
+    // DOF settings.
+    params.insert("f_stop", camera_node->GetLensApertureFNumber(time, FOREVER));
+    params.insert("focal_distance", camera_node->GetFocusDistance(time, FOREVER));
+    switch (camera_node->GetBokehShape(time, FOREVER))
+    {
+    case MaxSDK::IPhysicalCamera::BokehShape::Circular:
```
dictoon: Half-indent cases.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.h
```diff
@@ -44,7 +44,9 @@
 
 // Forward declarations.
 namespace renderer { class Camera; }
+namespace renderer { class ParamArray; }
 namespace renderer { class Project; }
+
```
dictoon: Remove this blank line, these are all forward declarations.

dictoon:
Regarding your question: What do you do exactly when you say "I've tried to pause the rendering"?

When the scene is edited in any way, there isn't much choice but _stopping_ rendering, reinitializing a bunch of things and restarting rendering, because entities may precompute and cache values based on their and other entities' parameters. So the general idea is to schedule a modification, return `ReinitializeRendering` from the renderer controller and apply modifications just before rendering starts again.

usakhelo:
By pause I meant setting `RendererController->set_status(Status::PauseRendering)`. I didn't investigate on why exactly it crashed, but I'll check and try to implement scheduling and proper updating.

dictoon:
Ah ok, `Status::PauseRendering` won't work, all it does is pause job execution, it doesn't do any reinitialization so entities may still hold a pointer to the old camera.

dictoon:
Nice!

PR #122, opened 2018-01-11
Add default lights when rendering material slots

### LINKED ISSUE #116: Don't turn off default lighting in Material Editor slots
This is the cause I thought blend doesn't blend light materials properly. Closing #107

usakhelo:
Small fix makes default lights always added when rendering material slots. Fixes #116 .

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -990,9 +990,10 @@ namespace
         const bool has_lights = !entities.m_lights.empty();
         const bool has_emitting_mats = has_light_emitting_materials(material_map);
         const bool has_emitting_env = !scene.get_environment()->get_parameters().get_optional<std::string>("environment_edf").empty();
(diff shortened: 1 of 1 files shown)


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -990,7 +990,8 @@ namespace
         const bool has_lights = !entities.m_lights.empty();
         const bool has_emitting_mats = has_light_emitting_materials(material_map);
         const bool has_emitting_env = !scene.get_environment()->get_parameters().get_optional<std::string>("environment_edf").empty();
-        if (!has_lights &&
+        if (rend_params.inMtlEdit ||
```
dictoon: Let's add parentheses to make the logic clearer:

    if (rend_params.inMtlEdit ||
        (!has_lights &&
         !has_emitting_mats &&
         !(has_emitting_env && settings.m_background_emits_light)))

dictoon:
Feel free to merge after adding the parentheses. Thanks!

PR #123, opened 2018-01-11
Update interactive rendering on viewport change, transforms of any viewport and changing properties of cameras

usakhelo:
Hi @dictoon ,
This pr changes the way camera is updating during interactive rendering.
I've made these changes more than a week ago and I don't remember things very clearly. I'll check it tomorrow one more time and then I'll ask you to review it.
Don't review it just yet!

### DIFF
--- a/src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
+++ b/src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
@@ -30,7 +30,6 @@
 #include "appleseedinteractive.h"
 
 // appleseed-max headers.
-#include "appleseedinteractive/interactiverenderercontroller.h"
 #include "appleseedinteractive/interactivesession.h"
 #include "appleseedinteractive/interactivetilecallback.h"
 #include "appleseedrenderer/appleseedrenderer.h"
@@ -277,7 +276,7 @@ namespace
                     
                 Matrix3 curr_mat;
                 vp13->GetAffineTM(curr_mat);
-                float curr_fov = vp13->GetFOV();
+                const float curr_fov = vp13->GetFOV();
 
                 if (m_last_mat != curr_mat ||
                     m_last_fov != curr_fov)
@@ -382,8 +381,7 @@ void AppleseedInteractiveRender::update_camera_object(INode* camera)
     get_view_params_from_view_node(view_params, camera, m_time);
 
     auto new_camera = build_camera(camera, view_params, m_bitmap, RendererSettings::defaults(), m_time);
-    get_render_session()->get_render_controller()->schedule_update(
-        std::unique_ptr<ScheduledAction>(new CameraObjectUpdateAction(m_project.ref(), new_camera)));
+    get_render_session()->schedule_camera_update(new_camera);
 }
 
 void AppleseedInteractiveRender::update_render_view()
@@ -405,8 +403,7 @@ void AppleseedInteractiveRender::update_render_view()
         get_view_params_from_viewport(view_params, view_exp, m_time);
 
     auto new_camera = build_camera(view_camera, view_params, m_bitmap, RendererSettings::defaults(), m_time);
-    get_render_session()->get_render_controller()->schedule_update(
-        std::unique_ptr<ScheduledAction>(new CameraObjectUpdateAction(m_project.ref(), new_camera)));
+    get_render_session()->schedule_camera_update(new_camera);
 
     if (view_camera != nullptr)
         m_node_callback.reset(new SceneChangeCallback(this, view_camera));
--- a/src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
+++ b/src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
@@ -37,6 +37,10 @@
 #include "appleseedrenderer/projectbuilder.h"
 #include "utilities.h"
 
+// Boost headers.
+#include "boost/thread/locks.hpp"
+#include "boost/thread/mutex.hpp"
+
 // 3ds Max headers.
 #include <assert1.h>
 #include <matrix3.h>
@@ -49,6 +53,9 @@ namespace asr = renderer;
 
 namespace
 {
+    boost::mutex                g_current_interactive_mutex;
+    AppleseedInteractiveRender* g_current_interactive;
+
     void get_view_params_from_viewport(
         ViewParams&             view_params,
         ViewExp&                view_exp,
@@ -190,25 +197,15 @@ namespace
           : m_renderer(renderer)
           , m_active_camera(camera)
         {
+            m_callback_key = GetISceneEventManager()->RegisterCallback(this, false, 100, true);
         }
 
-        virtual void ControllerOtherEvent(NodeKeyTab& nodes) override
+        ~SceneChangeCallback() override
         {
-            if (m_active_camera == nullptr || m_renderer == nullptr)
-                return;
-            
-            for (int i = 0, e = nodes.Count(); i < e; ++i)
-            {
-                if (NodeEventNamespace::GetNodeByKey(nodes[i]) == m_active_camera)
-                {
-                    m_renderer->update_camera_transform(m_active_camera);
-                    m_renderer->get_render_session()->reininitialize_render();
-                    break;
-                }
-            }
+            GetISceneEventManager()->UnRegisterCallback(m_callback_key);
         }
 
-        virtual void ModelOtherEvent(NodeKeyTab& nodes) override
+        void ModelOtherEvent(NodeKeyTab& nodes) override
         {
             if (m_active_camera == nullptr || m_renderer == nullptr)
                 return;
@@ -217,7 +214,7 @@ namespace
             {
                 if (NodeEventNamespace::GetNodeByKey(nodes[i]) == m_active_camera)
                 {
-                    m_renderer->update_camera_parameters(m_active_camera);
+                    m_renderer->update_camera_object(m_active_camera);
                     m_renderer->get_render_session()->reininitialize_render();
                     break;
                 }
@@ -225,8 +222,82 @@ namespace
         }
 
       private:
-        AppleseedInteractiveRender*     m_renderer;
-        INode*                          m_active_camera;
+        SceneEventNamespace::CallbackKey    m_callback_key;
+        AppleseedInteractiveRender*         m_renderer;
+        INode*                              m_active_camera;
+    };
+
+    class ViewportCallback 
+      : public RedrawViewsCallback
+    {
+      public:
+        ViewportCallback()
+          : m_current_view(nullptr)
+          , m_last_fov(0.0f)
+          , m_last_timer(0)
+          , m_max_hwnd(GetCOREInterface()->GetMAXHWnd())
+        {
+            m_last_mat.IdentityMatrix();
+
+            GetCOREInterface()->RegisterRedrawViewsCallback(this);
+        }
+
+        ~ViewportCallback() override
+        {
+            GetCOREInterface()->UnRegisterRedrawViewsCallback(this);
+        }
+
+        static VOID CALLBACK timer_proc(
+            _In_ HWND     hwnd,
+            _In_ UINT     msg,
+            _In_ UINT_PTR id,
+            _In_ DWORD    time
+        )
+        {
+            KillTimer(hwnd, id);
+            {
+                boost::mutex::scoped_lock lock(g_current_interactive_mutex);
+                if (g_current_interactive != nullptr)
+                {
+                    g_current_interactive->update_render_view();
+                    g_current_interactive->get_render_session()->reininitialize_render();
+                }
+            }
+        }
+
+        void proc(Interface* ip) override
+        {
+            ViewExp& view_exp = ip->GetActiveViewExp();
+            if (m_current_view == nullptr)
+                m_current_view = &view_exp;
+
+            if (view_exp.IsAlive())
+            {
+                ViewExp13* vp13 = reinterpret_cast<ViewExp13*>(view_exp.Execute(ViewExp::kEXECUTE_GET_VIEWEXP_13));
+                    
+                Matrix3 curr_mat;
+                vp13->GetAffineTM(curr_mat);
+                float curr_fov = vp13->GetFOV();
+
+                if (m_last_mat != curr_mat ||
+                    m_last_fov != curr_fov)
+                {
+                    m_last_mat = curr_mat;
+                    m_last_fov = curr_fov;
+                    if (m_last_timer == 0)
+                        m_last_timer = SetTimer(m_max_hwnd, 0, 100, timer_proc);
+                    else
+                        SetTimer(m_max_hwnd, m_last_timer, 100, timer_proc);
+                }
+            }
+        }
+
+      private:
+        ViewExp*    m_current_view;
+        float       m_last_fov;
+        Matrix3     m_last_mat;
+        UINT_PTR    m_last_timer;
+        HWND        m_max_hwnd;
     };
 }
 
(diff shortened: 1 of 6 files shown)


usakhelo:
This PR is ready for review.
Briefly what it does:
1. Registers viewport callback that is called each time viewport is re-drawn.
2. Adds scheduled action class to interactive rendering controller.
Callback checks if view matrix has changed since last call and if FOV has changed. If they have it schedules camera update.
SceneChangeCallback is left to only update camera parameters. Camera transforms update handled by viewport callback.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -190,43 +190,94 @@ namespace
           : m_renderer(renderer)
           , m_active_camera(camera)
         {
+            m_callback_key = GetISceneEventManager()->RegisterCallback(this, false, 100, true);
         }
 
-        virtual void ControllerOtherEvent(NodeKeyTab& nodes) override
+        virtual ~SceneChangeCallback()
```
dictoon: Even better:

    ~SceneChangeCallback() override

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -190,43 +190,94 @@ namespace
...
+        {
+            GetISceneEventManager()->UnRegisterCallback(m_callback_key);
+        }
+
+        virtual void ModelOtherEvent(NodeKeyTab& nodes) override
```
dictoon: This is clearly correct, but our convention is to omit `virtual` when a method is marked as `override` (since it implies `virtual` already).

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -190,43 +190,94 @@ namespace
...
+    };
+
+    class ViewportCallback 
+      : public RedrawViewsCallback
+    {
+      public:
+        ViewportCallback()
+          : m_current_view(nullptr)
+          , m_last_fov(0)
+          , m_last_timer(0)
```
dictoon: `nullptr`

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -190,43 +190,94 @@ namespace
...
+        INode*                              m_active_camera;
+    };
+
+    class ViewportCallback 
+      : public RedrawViewsCallback
+    {
+      public:
+        ViewportCallback()
+          : m_current_view(nullptr)
+          , m_last_fov(0)
```
dictoon: `0.0f`

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -190,43 +190,94 @@ namespace
...
-            for (int i = 0, e = nodes.Count(); i < e; ++i)
+            GetCOREInterface()->RegisterRedrawViewsCallback(this);
+        }
+
+        virtual ~ViewportCallback()
+        {
+            GetCOREInterface()->UnRegisterRedrawViewsCallback(this);
+        }
+
+        virtual void proc(Interface *ip) override
```
dictoon: `Interface* ip`

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -190,43 +190,94 @@ namespace
...
+                m_current_view = &view_exp;
+
+            if (view_exp.IsAlive())
             {
-                if (NodeEventNamespace::GetNodeByKey(nodes[i]) == m_active_camera)
+                ViewExp13* vp13 = reinterpret_cast<ViewExp13*>(view_exp.Execute(ViewExp::kEXECUTE_GET_VIEWEXP_13));
+                    
+                Matrix3 curr_mat;
+                vp13->GetAffineTM(curr_mat);
+                if (m_last_mat.IsIdentity())
```
dictoon: I find this strange. You're already updating `m_last_mat` and `m_last_fov` when you detect a change and create a timer below.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -190,43 +190,94 @@ namespace
...
+                    m_last_mat = curr_mat;
+
+                float curr_fov = vp13->GetFOV();
+                if (m_last_fov == 0)
```
dictoon: Same thing here.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -190,43 +190,94 @@ namespace
...
+
+                if (m_last_mat != curr_mat ||
+                    m_last_fov != curr_fov)
                 {
-                    m_renderer->update_camera_parameters(m_active_camera);
-                    m_renderer->get_render_session()->reininitialize_render();
-                    break;
+                    m_last_mat = curr_mat;
+                    m_last_fov = curr_fov;
+                    if (m_last_timer == 0)
```
dictoon: `nullptr`

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -305,39 +356,64 @@ asf::auto_release_ptr<asr::Project> AppleseedInteractiveRender::prepare_project(
...
+    if (view_camera != nullptr)
+        m_node_callback.reset(new SceneChangeCallback(this, view_camera));
 }
 
 InteractiveSession* AppleseedInteractiveRender::get_render_session()
 {
     return m_render_session.get();
 }
 
+AppleseedInteractiveRender* AppleseedInteractiveRender::m_current_interactive;
```
dictoon: Define before defining methods.

src/appleseed-max-impl/appleseedinteractive/interactiverenderercontroller.cpp
```diff
@@ -53,3 +58,8 @@ void InteractiveRendererController::set_status(const Status status)
 {
     m_status = status;
 }
+
+void InteractiveRendererController::schedule_update(std::unique_ptr<ScheduledAction> updater)
+{
+    m_scheduled_actions.push_back(std::move(updater));
```
dictoon: You need to include `<utility>` for `std::move()`.

src/appleseed-max-impl/appleseedinteractive/interactivesession.h
```diff
@@ -55,6 +55,8 @@ class InteractiveSession
     void reininitialize_render();
     void end_render();
 
+    InteractiveRendererController* get_render_controller();
+
   private:
     void render_thread();
```
dictoon: Make data members first, methods second.

src/appleseed-max-impl/appleseedinteractive/interactivesession.h
```diff
@@ -55,6 +55,8 @@ class InteractiveSession
     void reininitialize_render();
     void end_render();
 
+    InteractiveRendererController* get_render_controller();
```
dictoon: Make const, I think that should be possible.

src/appleseed-max-impl/appleseedinteractive/interactiverenderercontroller.h
```diff
@@ -42,6 +85,9 @@ class InteractiveRendererController
 
     void set_status(const Status status);
 
+    void schedule_update(std::unique_ptr<ScheduledAction> updater);
+
   private:
-    Status m_status;
+    std::vector<std::unique_ptr<ScheduledAction>>   m_scheduled_actions;
```
dictoon: Include `<vector>`.

src/appleseed-max-impl/appleseedinteractive/interactiverenderercontroller.h
```diff
@@ -30,6 +30,49 @@
...
+    }
+
+    virtual void update() override
+    {
+        m_project.get_scene()->cameras().clear();
+        m_project.get_scene()->cameras().insert(m_camera);
+    }
+
+  public:
+      foundation::auto_release_ptr<renderer::Camera>    m_camera;
```
dictoon: Fix indentation (should be four spaces in total).

src/appleseed-max-impl/appleseedinteractive/interactiverenderercontroller.h
```diff
@@ -30,6 +30,49 @@
...
+{
+  public:
+    virtual ~ScheduledAction() {}
+    virtual void update() = 0;
+};
+
+class CameraObjectUpdateAction : public ScheduledAction
+{
+  public:
+    explicit CameraObjectUpdateAction(
```
dictoon: You can remove `explicit` here since the argument takes two arguments. Again, it's not a mistake, just a minor styling thing.

src/appleseed-max-impl/appleseedinteractive/interactiverenderercontroller.h
```diff
@@ -30,6 +30,49 @@
 
 // appleseed.renderer headers.
 #include "renderer/api/rendering.h"
+#include "renderer/api/project.h"
```
dictoon: Order alphabetically.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -305,39 +356,64 @@ asf::auto_release_ptr<asr::Project> AppleseedInteractiveRender::prepare_project(
...
+
+VOID CALLBACK AppleseedInteractiveRender::timer_proc(
```
dictoon: I'm not sure why this method is defined here, to be honest. Why not define it in `ViewportCallback` which appears to be the only class using it?

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.h
```diff
@@ -97,14 +97,21 @@ class AppleseedInteractiveRender
...
     InteractiveSession* get_render_session();
+    static VOID CALLBACK timer_proc(
+        _In_ HWND     hwnd,
+        _In_ UINT     uMsg,
+        _In_ UINT_PTR idEvent,
+        _In_ DWORD    dwTime
+    );
 
   private:
+    static AppleseedInteractiveRender*              m_current_interactive;
```
dictoon: This is essentially a global variable. Since it appears to be accessed by multiple threads, it needs to be guarded (by a lock, typically) whenever it's read or modified.

src/appleseed-max-impl/appleseedinteractive/interactiverenderercontroller.cpp
```diff
@@ -41,6 +41,11 @@ InteractiveRendererController::InteractiveRendererController()
 
 void InteractiveRendererController::on_rendering_begin()
 {
+    for (auto& updater : m_scheduled_actions)
+    {
```
dictoon: You can remove the braces for one-liners.

src/appleseed-max-impl/appleseedinteractive/interactiverenderercontroller.h
```diff
@@ -30,6 +30,49 @@
...
+// appleseed.foundation headers.
+#include "foundation/utility/autoreleaseptr.h"
+
+// appleseed-max headers.
+#include "appleseedinteractive/appleseedinteractive.h"
+
+// Standard headers.
+#include <memory>
+
+namespace renderer { class Camera; }
```
dictoon: Add the `// Forward declarations.` marker.

src/appleseed-max-impl/appleseedinteractive/interactiverenderercontroller.h
```diff
@@ -30,6 +30,49 @@
...
+namespace renderer { class Project; }
+
+class ScheduledAction
+{
+  public:
+    virtual ~ScheduledAction() {}
+    virtual void update() = 0;
+};
+
+class CameraObjectUpdateAction : public ScheduledAction
```
dictoon:
```
class CameraObjectUpdateAction
  : public ScheduledAction
```

usakhelo:
Thanks for review @dictoon.
Couple of questions (this the email so I can't comment specific comments):

1. So I can drop virtual keyword in derived class completely and only use
override?

2. How should project be set up so that VS doesn't find some of headers by
itself? Does it worth doing that? I think it would be safer and cleaner if
VS doesn't find some headers for you.

3. Thanks for tip on explicit.

4. To guard global variable lock should also be global, right? Is it safe
to access global lock?

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
usakhelo: It doesn't accept nullptr - says nullptr cannot be used to initialize UINT_PTR

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
usakhelo: I thought they should be initialized first to avoid first changed just because _last_ vars are uninitialized. But it's probably not big deal, I'll remove this part.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
usakhelo: Moved.

src/appleseed-max-impl/appleseedinteractive/interactivesession.h
usakhelo: It doesn't compile with const - it calls `InteractiveRendererController::schedule_update(std::unique_ptr<ScheduledAction> updater)` on this pointer.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.h
usakhelo: Added global lock the same way as it's done in logtarget.cpp

usakhelo: And moved `m_current_interactive` to be really global variable, not static member: `g_current_interactive`

dictoon:
> 1. So I can drop virtual keyword in derived class completely and only use
override?

Yes. Whenever you override a virtual method, you should mark it as `override` and drop the `virtual` keyword.

> 2. How should project be set up so that VS doesn't find some of headers by
itself? Does it worth doing that? I think it would be safer and cleaner if
VS doesn't find some headers for you.

Unfortunately I'm not aware of any easy solution. We just need to be diligent and include what we think is required, but of course we're all making tons of mistakes. It's not a big deal. Let's just add/remove header files when we notice something wrong.

Note that Google has released a tool called iwyu ("Include What You Use") which reports all violations (missing or redundant headers). It's only available on Linux I think, although I seem to remember seeing some Windows builds floating around.

I've run iwyu on appleseed and it reported _thousands_ of violations, with many false positives since the problem is actually extremely difficult to solve. So it's not perfect either.

Let's just not sweat too much about it.

> 3. Thanks for tip on explicit.

Note that your code was very much valid and I could imagine marking absolutely all constructors as explicit by default being a reasonable style guideline in some projects. I just wish C++ had adopted explicit-by-default (with an `implicit` keyword) than the other way around.

> 4. To guard global variable lock should also be global, right? Is it safe
to access global lock?

Yes, static. It's safe to access a lock from multiple threads since it's its very purpose. You can grep the appleseed codebase for `mutex` or `lock` and you'll find plenty of usage patterns.

dictoon:
I checked and your usage of `mutex` and `scoped_lock` is perfect.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -305,32 +376,40 @@ asf::auto_release_ptr<asr::Project> AppleseedInteractiveRender::prepare_project(
...
-#if MAX_RELEASE >= 18000
-    MaxSDK::IPhysicalCamera* phys_camera = dynamic_cast<MaxSDK::IPhysicalCamera*>(camera->EvalWorldState(m_time).obj);
-
-    if (phys_camera && phys_camera->GetDOFEnabled(m_time, FOREVER))
-    {
-        set_camera_dof_params(params, phys_camera, m_bitmap, m_time);
-    }
-#endif
+    auto new_camera = build_camera(camera, view_params, m_bitmap, RendererSettings::defaults(), m_time);
+    get_render_session()->get_render_controller()->schedule_update(
```
dictoon: I would suggest getting rid of `get_render_session()` and using `m_render_session` instead. The less stuff we expose, the clearer the code.

dictoon: I think I would go even further: I would not expose the renderer controller via `get_render_controller()`, but instead move `schedule_update()` to the render session.

usakhelo: Moved.

src/appleseed-max-impl/appleseedinteractive/appleseedinteractive.cpp
```diff
@@ -190,43 +196,107 @@ namespace
...
+            if (m_current_view == nullptr)
+                m_current_view = &view_exp;
+
+            if (view_exp.IsAlive())
+            {
+                ViewExp13* vp13 = reinterpret_cast<ViewExp13*>(view_exp.Execute(ViewExp::kEXECUTE_GET_VIEWEXP_13));
+                    
+                Matrix3 curr_mat;
+                vp13->GetAffineTM(curr_mat);
+                float curr_fov = vp13->GetFOV();
```
dictoon: Make const.

src/appleseed-max-impl/appleseedinteractive/interactivesession.cpp
```diff
@@ -93,3 +92,11 @@ void InteractiveSession::end_render()
     if (m_render_thread.joinable())
         m_render_thread.join();
 }
+
+void InteractiveSession::schedule_camera_update(
+    asf::auto_release_ptr<asr::Camera>  camera
+)
```
dictoon: Move closing parenthesis to previous line.

src/appleseed-max-impl/appleseedinteractive/interactivesession.h
```diff
@@ -29,17 +29,22 @@
...
+#include "foundation/utility/autoreleaseptr.h"
+
 // Standard headers.
 #include <memory>
 #include <thread>
 
 // Forward declarations.
+namespace renderer { class Camera; }
 namespace renderer { class Project; }
+
```
dictoon: Remove blank line.

dictoon:
One last thing: Could you rebase the PR, merge "comments fix" into the previous commit, and capitalize the first word of the second commit please? Feel free to merge after that. Thank you!

usakhelo:
Done, merging.
Thanks for review!

PR #127, opened 2018-02-02
Add SSS Set id parameter to Object Properties Modifier

usakhelo:
Hi,

this PR adds SSS Set id parameter to Obj. Properties modifier. Also I added "appleseed" in the modifier title because by default Max doesn't show modifier categories in the list, so it is not clear what `Object Properties` modifier means.

Thanks for review,
Sergo.


dictoon:
Thank you!

PR #129, opened 2018-02-10
Support closure2surface shader, fix obj. properties, support Substance materials

usakhelo:
- Support of changes to 3ds max shaders made in appleseed [PR 1842](https://github.com/appleseedhq/appleseed/pull/1842)
- Fixes Object Properties modifier errors when SSS Set Id field is empty
- Adds support of Substance maps to "Use Procedural" mode


dictoon:
Thanks!

PR #135, opened 2018-03-03
OSL Plugin fixes

usakhelo:
Hi,
Fixes in this PR:

1. Shader files in separate directories
2. Shaders from env. variable path override shaders in plugin directory
3. Texture buttons in OSL material plugins are update properly
4. Parameters are hidden based on special metadata token
5. Vector parameter connection works
6. Viewport shading for OSL materials

Thanks for review.

### DIFF
--- a/src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
+++ b/src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
@@ -605,6 +605,12 @@ asf::auto_release_ptr<asr::Material> OSLMaterial::create_osl_material(
                 break;
 
               case MaxParam::String:
+                {
+                    const wchar_t* str_value;
+                    GetParamBlock(0)->GetValue(max_param.m_max_param_id, t, str_value, FOREVER);
+                    if (str_value != nullptr)
+                        params.insert(max_param.m_osl_param_name.c_str(), fmt_osl_expr(wide_to_utf8(str_value)));
+                }
                 break;
             }
         }
--- a/src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
+++ b/src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
@@ -77,6 +77,7 @@ Class_ID OSLMaterial::get_class_id()
 
 OSLMaterial::OSLMaterial(Class_ID class_id, OSLPluginClassDesc* class_desc)
   : m_pblock(nullptr)
+  , m_bump_pblock(nullptr)
   , m_classid(class_id)
   , m_class_desc(class_desc)
   , m_shader_info(nullptr)
@@ -142,33 +143,50 @@ int OSLMaterial::SubNumToRefNum(int subNum)
 
 int OSLMaterial::NumParamBlocks()
 {
-    return 1;
+    return m_has_bump_params ? 2 : 1;
 }
 
 IParamBlock2* OSLMaterial::GetParamBlock(int i)
 {
-    return i == 0 ? m_pblock : nullptr;
+    if (i == 0)
+        return m_pblock;
+    else if (i == 1)
+        return m_bump_pblock;
+    else
+        return nullptr;
 }
 
 IParamBlock2* OSLMaterial::GetParamBlockByID(BlockID id)
 {
-    return id == m_pblock->ID() ? m_pblock : nullptr;
+    if (id == m_pblock->ID())
+        return m_pblock;
+    else if (id == m_bump_pblock->ID())
+        return m_bump_pblock;
+    else
+        return nullptr;
 }
 
 int OSLMaterial::NumRefs()
 {
-    return 1;
+    return m_has_bump_params ? 2 : 1;
 }
 
 RefTargetHandle OSLMaterial::GetReference(int i)
 {
-    return i == 0 ? m_pblock : nullptr;
+    if (i == 0)
+        return m_pblock;
+    else if (i == 1)
+        return m_bump_pblock;
+    else
+        return nullptr;
 }
 
 void OSLMaterial::SetReference(int i, RefTargetHandle rt_arg)
 {
     if (i == 0)
         m_pblock = static_cast<IParamBlock2*>(rt_arg);
+    else if (i == 1)
+        m_bump_pblock = static_cast<IParamBlock2*>(rt_arg);
 }
 
 RefResult OSLMaterial::NotifyRefChanged(
@@ -188,6 +206,8 @@ RefResult OSLMaterial::NotifyRefChanged(
       case REFMSG_CHANGE:
           if (hTarget == m_pblock)
               m_class_desc->GetParamBlockDesc(0)->InvalidateUI(m_pblock->LastNotifyParamID());
+          else if (hTarget == m_bump_pblock)
+              m_class_desc->GetParamBlockDesc(1)->InvalidateUI(m_bump_pblock->LastNotifyParamID());
         break;
     }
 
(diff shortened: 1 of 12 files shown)


src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
```diff
@@ -47,6 +48,7 @@
 
 // 3ds Max headers.
 #include <iparamm2.h>
+#include <hsv.h>
```
dictoon: Order alphabetically.

src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
```diff
@@ -224,17 +254,25 @@ void OSLMaterial::SetSubTexmap(int i, Texmap* tex_map)
...
-        m_pblock->SetValue(param_id, 0, tex_map);
-
-        if (m_pblock->GetParameterType(param_id - 1) == TYPE_POINT3)
+        const int param_id = m_texture_id_map[i].first;
+        if (m_has_bump_params && i == m_texture_id_map.size() - 1)
         {
-            IParamMap2* map = m_pblock->GetMap(0);
-            if (map != nullptr)
-                map->SetText(param_id, tex_map == nullptr ? L"" : L"M");
+            BOOL result = m_bump_pblock->SetValue(param_id, 0, tex_map);
```
dictoon: Remove `result` variable (unused).

src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
```diff
@@ -224,17 +254,25 @@ void OSLMaterial::SetSubTexmap(int i, Texmap* tex_map)
...
+            m_class_desc->GetParamBlockDesc(1)->InvalidateUI(param_id);
         }
+        else
+        {
+            BOOL result = m_pblock->SetValue(param_id, 0, tex_map);
```
dictoon: Remove `result`.

src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
```diff
@@ -347,27 +389,33 @@ Color OSLMaterial::GetAmbient(int mtlNum, BOOL backFace)
 
 Color OSLMaterial::GetDiffuse(int mtlNum, BOOL backFace)
 {
-    return Color(0.0f, 0.0f, 0.0f);
+    Mtl* mtl = nullptr;
+    Interval iv;
+    if (m_submaterial_map.size() > 0)
```
dictoon: if (!m_submaterial_map.empty())

is better.

src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
```diff
@@ -347,27 +389,33 @@ Color OSLMaterial::GetAmbient(int mtlNum, BOOL backFace)
...
 float OSLMaterial::GetShinStr(int mtlNum, BOOL backFace)
 {
-    return 0.0f;
+    return get_viewport_shininess_amount();
 }
 
 float OSLMaterial::GetXParency(int mtlNum, BOOL backFace)
 {
-    return 0.0f;
+    float transparency = get_viewport_transparency_amount();
```
dictoon: Make const.

src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
```diff
@@ -132,59 +132,76 @@ void OSLParamDlg::add_ui_parameter(
...
-
-      case MaxParam::String:
-        dialog_template.AddComponent((LPCSTR)"CustEdit", (LPCSTR)"Edit", WS_VISIBLE | WS_TABSTOP, NULL, Col2X, y_pos, TexButtonWidth, EditHeight, ctrl_id++);
-        break;
-
-      case MaxParam::Closure:
-        dialog_template.AddComponent((LPCSTR)"CustButton", (LPCSTR)"", WS_VISIBLE, NULL, Col3X, y_pos, TexButtonWidth, EditHeight, ctrl_id++);
+        switch (max_param.m_param_type)
+        {
+        case MaxParam::Float:
```
dictoon: Half-indent `case` statements.

src/appleseed-max-impl/appleseedoslplugin/oslshadermetadata.cpp
```diff
@@ -102,6 +102,7 @@ namespace
 OSLParamInfo::OSLParamInfo(const asf::Dictionary& paramInfo)
   : hasDefault(false)
   , divider(false)
+  , lockGeom(true)
```
dictoon: I think we should change `OSLParamInfo` to use the same naming conventions as the rest of the project (`m_` prefix and snake_case for member variables).

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
```diff
@@ -74,28 +103,28 @@ namespace
...
-            if (tex_map != nullptr)
-                map->SetText(id, is_short ? L"M" : tex_map->GetFullName());
-            else
-                map->SetText(id, is_short ? L"" : L"None");
+            set_button_text(pblock, id);
         }
     };
     
+    class BumpTextureAccessor
+        : public PBAccessor
```
dictoon: Only half-indentation here.

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
```diff
@@ -74,28 +103,28 @@ namespace
...
+    {
+    public:
```
dictoon: Half-indent.

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.h
```diff
@@ -46,18 +47,26 @@ class OSLShaderRegistry
...
     int get_size() const;
     void create_class_descriptors();
-    void add_parameter(
+    void add_const_parameter(
         ParamBlockDesc2*        pb_desc,
         const OSLParamInfo&     osl_param,
         MaxParam&               param_info,
         IdNameMap&              string_map,
+        const int               param_id,
+        int&                    ctrl_id,
```
dictoon: I find it a bit surprising that `add_const_parameter()` needs to modify `ctrl_id` and `string_id` but `add_input_parameter()` don't. Is there a fundamental reason why?

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.h
```diff
@@ -37,6 +37,7 @@
 
 // Standard headers.
 #include <vector>
+#include <string>
```
dictoon: Order alphabetically.

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.h
usakhelo: Yes.. I've tried to make `add_parameter()` prettier by splitting in two functions and moving some increments outside.
Thing is that `add_const_parameter()` creates variable number of controls (like three point3 spinners) and string objects (for comboboxes).
`add_input_parameter()` add fixed number of controls, that's why it doesn't change ctrl_id from inside the function body.

usakhelo:
@dictoon I've fixed the comments and add one more commit where broken output parameter is fixed and string inputs connected properly. That makes as_texture shader work.

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.h
dictoon: Ok, thanks for the explanation.

PR #139, opened 2018-03-14
Reset spinner to the default value, accept output map in bump

usakhelo:
Hi @dictoon 
Small PR with two fixes:
1. Reset spinners of OSL plugins to the default value on right click
2. Bump map accepts Output map (and other "supported" textures in future)

Thanks,
Sergo.


dictoon:
Nice, thanks!

This commit title doesn't make sense to me: "Supported texture as a bump and normal map", could you clarify?

usakhelo:
Hi @dictoon ,
I renamed the commit to "Accept Output map as a bump map"

dictoon:
Thanks! Merging.

PR #140, opened 2018-03-16
Output Selector plugin

usakhelo:
Hi @dictoon 

This PR adds Output Selector plugin for OSL Textures. Now in order to connect OSL texture to an input user can choose an output from texture's UI or user can add this plugin and choose output in the plugin's UI.
This basically compensates lack of multiple outputs on OSL textures.

Thanks,
Sergo.

### DIFF
--- a/src/appleseed-max-impl/appleseedoslplugin/osltexture.h
+++ b/src/appleseed-max-impl/appleseedoslplugin/osltexture.h
@@ -41,6 +41,10 @@
 #include <maxtypes.h>
 #undef base_type
 
+// Standard headers.
+#include <string>
+#include <vector>
+
 // Forward declarations.
 namespace renderer { class ShaderGroup; }
 class OSLPluginClassDesc;
@@ -105,7 +109,15 @@ class OSLTexture
     void create_osl_texture(
         renderer::ShaderGroup&  shader_group,
         const char*             material_node_name,
+        const char*             material_input_name,
+        const int               output_slot_index);
+
+    void create_osl_texture(
+        renderer::ShaderGroup&  shader_group,
+        const char*             material_node_name,
         const char*             material_input_name);
+
+    std::vector<std::string> get_output_names() const;
 
   protected:
     virtual void SetReference(int i, RefTargetHandle rtarg) override;
--- a/src/appleseed-max-impl/appleseedoslplugin/osltexture.h
+++ b/src/appleseed-max-impl/appleseedoslplugin/osltexture.h
@@ -105,7 +105,7 @@ class OSLTexture
     Point3 EvalNormalPerturb(ShadeContext& sc) override;
     UVGen* GetTheUVGen() override;
 
-    // Connect output to a material.
+    // Create OSL shader and connect to a given output.
     void create_osl_texture(
         renderer::ShaderGroup&  shader_group,
         const char*             material_node_name,
--- a/src/appleseed-max-impl/builtinmapsupport.cpp
+++ b/src/appleseed-max-impl/builtinmapsupport.cpp
@@ -54,15 +54,16 @@ void connect_output_selector(
 
     Texmap* input_map = nullptr;
     get_paramblock_value_by_name(texmap->GetParamBlock(0), L"source_map", t, input_map, FOREVER);
-    int output_index = 0;
-    get_paramblock_value_by_name(texmap->GetParamBlock(0), L"output_index", t, output_index, FOREVER);
-
     if (input_map != nullptr && is_osl_texture(input_map))
     {
         OSLTexture* osl_texture = static_cast<OSLTexture*>(input_map);
         auto output_names = osl_texture->get_output_names();
 
-        if (output_index > 0 && --output_index < output_names.size())
+        int output_index = 0;
+        get_paramblock_value_by_name(texmap->GetParamBlock(0), L"output_index", t, output_index, FOREVER);
+        DbgAssert(output_index >= 1);
+        output_index -= 1;
+        if (output_index < output_names.size())
         {
             osl_texture->create_osl_texture(
                 shader_group,
--- a/src/appleseed-max-impl/builtinmapsupport.cpp
+++ b/src/appleseed-max-impl/builtinmapsupport.cpp
@@ -62,7 +62,7 @@ void connect_output_selector(
         OSLTexture* osl_texture = static_cast<OSLTexture*>(input_map);
         auto output_names = osl_texture->get_output_names();
 
-        if (--output_index < output_names.size())
+        if (output_index > 0 && --output_index < output_names.size())
         {
             osl_texture->create_osl_texture(
                 shader_group,
@@ -74,7 +74,7 @@ void connect_output_selector(
 }
 
 void connect_output_map(
-    asr::ShaderGroup&  shader_group,
+    asr::ShaderGroup&   shader_group,
     const char*         material_node_name,
     const char*         material_input_name,
     Texmap*             texmap,
--- a/src/appleseed-max-impl/builtinmapsupport.cpp
+++ b/src/appleseed-max-impl/builtinmapsupport.cpp
@@ -62,11 +62,14 @@ void connect_output_selector(
         OSLTexture* osl_texture = static_cast<OSLTexture*>(input_map);
         auto output_names = osl_texture->get_output_names();
 
-        osl_texture->create_osl_texture(
-            shader_group,
-            material_node_name,
-            material_input_name,
-            static_cast<int>((output_index - 1) % output_names.size()));
+        if (--output_index < output_names.size())
+        {
+            osl_texture->create_osl_texture(
+                shader_group,
+                material_node_name,
+                material_input_name,
+                output_index);
+        }
     }
 }
 
--- a/src/appleseed-max-impl/builtinmapsupport.cpp
+++ b/src/appleseed-max-impl/builtinmapsupport.cpp
@@ -30,6 +30,7 @@
 #include "builtinmapsupport.h"
 
 // appleseed-max headers.
+#include "appleseedoslplugin/osltexture.h"
 #include "oslutils.h"
 #include "utilities.h"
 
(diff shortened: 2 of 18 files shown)


src/appleseed-max-impl/appleseedoslplugin/osltexture.h
```diff
@@ -102,11 +106,19 @@ class OSLTexture
     UVGen* GetTheUVGen() override;
 
     // Connect output to a material.
```
dictoon: Is this comment up-to-date?

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -43,6 +44,47 @@
 namespace asf = foundation;
 namespace asr = renderer;
 
+void connect_output_selector(
+    renderer::ShaderGroup&  shader_group,
```
dictoon: Maybe `asr::ShaderGroup`?

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -43,6 +44,47 @@
...
+
+    if (input_map != nullptr && is_osl_texture(input_map))
+    {
+        auto output_names = static_cast<OSLTexture*>(input_map)->get_output_names();
+
+        static_cast<OSLTexture*>(input_map)->create_osl_texture(
+            shader_group,
+            material_node_name,
+            material_input_name,
+            static_cast<int>((output_index - 1) % output_names.size()));
```
dictoon: What is the purpose of the modulo operation here? It looks suspicious :)

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -43,6 +44,47 @@
...
+    Texmap* input_map = nullptr;
+    get_paramblock_value_by_name(texmap->GetParamBlock(0), L"source_map", t, input_map, FOREVER);
+    int output_index = 0;
+    get_paramblock_value_by_name(texmap->GetParamBlock(0), L"output_index", t, output_index, FOREVER);
+
+    if (input_map != nullptr && is_osl_texture(input_map))
+    {
+        auto output_names = static_cast<OSLTexture*>(input_map)->get_output_names();
+
+        static_cast<OSLTexture*>(input_map)->create_osl_texture(
```
dictoon: Cast once.

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -43,6 +44,47 @@
...
+            shader_group,
+            material_node_name,
+            material_input_name,
+            static_cast<int>((output_index - 1) % output_names.size()));
+    }
+}
+
+void connect_output_selector(
+    renderer::ShaderGroup&  shader_group,
```
dictoon: Use `asr` prefix.

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -43,6 +44,47 @@
...
+    const char*             material_input_name,
+    Texmap*                 texmap,
+    const float             const_value)
+{
+    connect_output_selector(
+        shader_group,
+        material_node_name,
+        material_input_name,
+        texmap,
+        Color());
```
dictoon: Is this correct? Shouldn't it be `Color(const_value)`? Just checking...

src/appleseed-max-impl/builtinmapsupport.h
```diff
@@ -68,6 +85,17 @@ void create_supported_texture(
     auto part_a = texmap->ClassID().PartA();
```
dictoon: Move these declarations lower in the function, where they're first needed.

src/appleseed-max-impl/osloutputselectormap/osloutputselector.cpp
```diff
@@ -0,0 +1,604 @@
...
+        ParamBlockIdOutputSelector,                 // parameter block's ID
+        L"appleseedOuputSelectorParams",            // internal parameter block's name
+        0,                                          // ID of the localized name string
+        &g_appleseed_outputselector_classdesc,      // class descriptor
+        P_AUTO_CONSTRUCT + P_AUTO_UI,               // block flags
+
+         // --- P_AUTO_CONSTRUCT arguments ---
+        ParamBlockRefOutputSelector,                // parameter block's reference number
+
+        // --- P_AUTO_UI arguments for Parameters rollup ---
```
dictoon: Add a blank line after this comment since it applies not only to the next block, but to all following blocks.

src/appleseed-max-impl/osloutputselectormap/osloutputselector.cpp
```diff
@@ -0,0 +1,604 @@
...
+        param_map->GetParamBlock()->GetValue(ParamIdSourceMap, time, texmap, FOREVER);
+
+        if (texmap != nullptr && is_osl_texture(texmap))
+        {
+            auto output_names = static_cast<OSLTexture*>(texmap)->get_output_names();
+            for (const auto& name : output_names)
+                SendMessage(GetDlgItem(dlg_hwnd, IDC_COMBO_OUTPUT_NAME), CB_ADDSTRING, 0, reinterpret_cast<LPARAM>(utf8_to_wide(name).c_str()));
+
+            int output_index = param_map->GetParamBlock()->GetInt(ParamIdOutputIndex, time, FOREVER);
+            SendMessage(GetDlgItem(dlg_hwnd, IDC_COMBO_OUTPUT_NAME), CB_SETCURSEL, (output_index - 1) % output_names.size(), 0);
```
dictoon: Same question regarding the modulo.

src/appleseed-max-impl/osloutputselectormap/osloutputselector.cpp
```diff
@@ -0,0 +1,604 @@
...
+        }
+        else
+        {
+            SendMessage(GetDlgItem(dlg_hwnd, IDC_COMBO_OUTPUT_NAME), CB_RESETCONTENT, 0, 0);
+        }
+    }
+
+    class OutputSelectorDlgProc
+        : public ParamMap2UserDlgProc
```
dictoon: Half-indent.

src/appleseed-max-impl/osloutputselectormap/osloutputselector.cpp
```diff
@@ -0,0 +1,604 @@
...
+    {
+    public:
```
dictoon: Half-indent.

src/appleseed-max-impl/builtinmapsupport.cpp
usakhelo: Last parameter is discarded in both version so override is not needed here.. Thanks for pointing that out!

src/appleseed-max-impl/builtinmapsupport.cpp
usakhelo: `output_index` should not be more than `output_names.size()` but instead of constraining the parameter I thought I'll use modulo so that any value can be used as an index for the names vector.
I've seen similar behavior somewhere else in Max. I'll try to specify in slack..

usakhelo: Material IDs of the meshes work in the same way: let's say we have mesh with face id-s 1, 2, 3. I can apply multi-material with 4 sub-materials to that mesh. 4th sub-material will be applied to the face with ID = 1.
Even when some index parameter has value bigger than collection it indexes, instead of discarding bigger values they are still used.
Hope it's clear.

usakhelo:
I'm adding new commit but once PR is approved I'll squash and rebase.

dictoon:
> I'm adding new commit but once PR is approved I'll squash and rebase.

Perfect, that's the best way to do it.

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -43,12 +44,41 @@
...
+    get_paramblock_value_by_name(texmap->GetParamBlock(0), L"source_map", t, input_map, FOREVER);
+    int output_index = 0;
+    get_paramblock_value_by_name(texmap->GetParamBlock(0), L"output_index", t, output_index, FOREVER);
+
+    if (input_map != nullptr && is_osl_texture(input_map))
+    {
+        OSLTexture* osl_texture = static_cast<OSLTexture*>(input_map);
+        auto output_names = osl_texture->get_output_names();
+
+        if (--output_index < output_names.size())
```
dictoon: That doesn't look correct. What if `output_index` is 0?

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -43,12 +44,41 @@
...
+    }
+}
+
 void connect_output_map(
-    renderer::ShaderGroup&  shader_group,
-    const char*             material_node_name,
-    const char*             material_input_name,
-    Texmap*                 texmap,
-    const Color             const_value)
+    asr::ShaderGroup&  shader_group,
```
dictoon: Fix indentation.

src/appleseed-max-impl/osloutputselectormap/osloutputselector.cpp
```diff
@@ -0,0 +1,616 @@
...
+
+    return result;
+}
+
+AColor OSLOutputSelector::EvalColor(ShadeContext& sc)
+{
+    return Color(0.13f, 0.58f, 1.0f);
+}
+
+Point3 OSLOutputSelector::EvalNormalPerturb(ShadeContext& /*sc*/)
```
dictoon: `sc` is commented in this method but not in the one above.

src/appleseed-max-impl/plugin.cpp
```diff
@@ -38,6 +38,7 @@
 #include "appleseedplasticmtl/appleseedplasticmtl.h"
 #include "appleseedrenderer/appleseedrenderer.h"
 #include "appleseedsssmtl/appleseedsssmtl.h"
+#include "osloutputselectormap/osloutputselector.h"
```
dictoon: Order alphabetically.

src/appleseed-max-impl/utilities.cpp
```diff
@@ -31,6 +31,7 @@
 
 // appleseed-max headers.
 #include "appleseedoslplugin/osltexture.h"
+#include "osloutputselectormap/osloutputselector.h"
```
dictoon: Order alphabetically.

src/appleseed-max-impl/builtinmapsupport.cpp
usakhelo: That parameter should not be less than 1 but I agree, it's not safe.
Would this kind of check work? `if (output_index > 0 && --output_index < output_names.size())`

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -62,7 +62,7 @@ void connect_output_selector(
         OSLTexture* osl_texture = static_cast<OSLTexture*>(input_map);
         auto output_names = osl_texture->get_output_names();
 
-        if (--output_index < output_names.size())
+        if (output_index > 0 && --output_index < output_names.size())
```
dictoon: It's still very unusual and somewhat dangerous to modify `output_index` in the `if` statement.

If we _know_ for sure that `output_index` is always >= 1, then what I'd do in that case is assert (`DbgAssert()` in Max), and then decrease it.

    DbgAssert(output_index >= 1);
    output_index -= 1;

    if (output_index < output_names.size())
    {
        ...
    }

Why is `output_index` always >= 1?

usakhelo:
Yes, I agree it doesn't look good. Reason why it should never be less than
1 is that parameter declaration contains range tag: `p_range, 1, 100`. I'm
not sure how bullet proof this declaration is. May be it should be in
another `if(output_index > 0)` statement?

dictoon:
As long as we control the range in the code, then it's fine and an assertion is enough.

dictoon:
Thank you, merged!

PR #143, opened 2018-03-22
Make UI groups nested according to shader metadata

usakhelo:
Hi @dictoon 
This PR makes nested UI groups based on shader metadata. Groups are named with short name without periods or bars.

Thanks for review!
Sergo.

### DIFF
--- a/src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
+++ b/src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
@@ -100,7 +100,7 @@ namespace
         for (auto child : page->m_children)
             adjust_group_heights(child);
 
-        page->m_height += 12;                                   // extra space for the contols
+        page->m_height += 12;                                   // extra space for the controls
         if (page->m_parent != nullptr)
         {
             page->m_parent->m_height += page->m_height;
@@ -110,15 +110,15 @@ namespace
 
     void adjust_group_positions(PageGroup* page)
     {
-        int y_pos = page->m_y_pos + page->m_height;
+        int y_pos = page->m_y + page->m_height;
         for (auto it = page->m_children.rbegin(); it != page->m_children.rend(); ++it)
         {
             PageGroup* child = *it;
-            child->m_y_pos = y_pos - child->m_height - 3;       // spacing between the groups
-            y_pos = child->m_y_pos;
+            child->m_y = y_pos - child->m_height - 3;       // spacing between the groups
+            y_pos = child->m_y;
 
             child->m_width = page->m_width - 8;
-            child->m_x_pos = page->m_x_pos + 4;
+            child->m_x = page->m_x + 4;
         }
 
         for (auto child : page->m_children)
@@ -149,7 +149,7 @@ namespace
         for (auto& page : pages)
         {
             auto parent_name = page.first;
-            size_t dot_pos = parent_name.find_last_of('.');
+            const size_t dot_pos = parent_name.find_last_of('.');
             if (dot_pos == std::string::npos)
             {
                 page.second.m_short_name = page.first;
@@ -173,6 +173,9 @@ namespace
 
         adjust_group_heights(&root_page);
         adjust_group_positions(&root_page);
+
+        for (auto page : root_page.m_children)
+            page->m_parent = nullptr;
 
         return root_page.m_height;
     }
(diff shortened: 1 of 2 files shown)


src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
```diff
@@ -94,6 +94,88 @@ namespace
             }
         }
     };
+
+    void adjust_group_heights(PageGroup* page)
+    {
+        for (auto child : page->m_children)
+            adjust_group_heights(child);
+
+        page->m_height += 12;                                   // extra space for the contols
```
dictoon: Typo: *controls*

src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
```diff
@@ -94,6 +94,88 @@ namespace
...
+        }
+        
+        // Build page tree.
+        PageGroup root_page;
+        root_page.m_width = 217;
+        std::vector<PageGroup*> page_tree;
+        for (auto& page : pages)
+        {
+            auto parent_name = page.first;
+            size_t dot_pos = parent_name.find_last_of('.');
```
dictoon: Make const.

src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
```diff
@@ -94,6 +94,88 @@ namespace
...
+            if (dot_pos == std::string::npos)
+            {
+                page.second.m_short_name = page.first;
+                page_tree.push_back(&page.second);
+                page.second.m_parent = &root_page;
```
dictoon: You're taking a reference to a local variable. Is that OK?

src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.h
```diff
@@ -55,8 +56,21 @@ class OSLShaderInfo;
...
-    int height;
+    PageGroup()
+        : m_y_pos(0)
+        , m_x_pos(0)
+        , m_height(0)
+        , m_width(0)
+        , m_parent(nullptr)
+    {}
+
+    int                         m_x_pos;
```
dictoon: Maybe just `m_x` and `m_y`?

src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
usakhelo: I should reset m_parent to nullptr before returning from function probably.

src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
```diff
@@ -94,6 +94,91 @@ namespace
...
+            {
+                const int param_height = 
+                    param.m_max_param.m_param_type == MaxParam::IntMapper ||
+                    param.m_max_param.m_param_type == MaxParam::StringPopup ? 15 : 12;
+
+                pages[param.m_page].m_height += param_height;
+            }
+        }
+        
+        // Build page tree.
```
dictoon: In fact, more precisely this block of code builds the _root page_ rather than the page tree, no?

src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
```diff
@@ -94,6 +94,91 @@ namespace
...
+        PageGroup root_page;
+        root_page.m_width = 217;
+        std::vector<PageGroup*> page_tree;
```
dictoon: Would it be possible to eliminate this variable and instead use directly `root_page.m_children`? It would make things clearer: not only is `page_tree` not a page _tree_ but just a page _vector_, but also it would make it clear that the first step is only about collecting the root page's children.

src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
```diff
@@ -175,14 +265,18 @@ void OSLParamDlg::add_ui_parameter(
     const int EditHeight = 10;
     const int TexButtonWidth = 84;
 
+    int col_1x = 10;
```
dictoon: May I suggest `col1_x` (and then use this convention everywhere)? This would make it clearer that it's the X position of col1, because 1x means "one time" to me...

src/appleseed-max-impl/appleseedoslplugin/oslparamdlg.cpp
usakhelo: There's no tree before this block so I thought I'm building a tree here :) No problem, I'll rename.

PR #154, opened 2018-04-06
Add RenderElement plugin to support appleseed's AOVs

usakhelo:
Hi @dictoon 
This is render element plugin that allows users to render appleseed AOVs.
Plugin basically used to store AOV type and output filename.
I could not find a way to copy AOV image from appliseed image to 3ds max render element bitmap, and for that reason at some stage render elements are turned off and files are saved with appleseed's function `project->get_frame()->write_main_and_aov_images();`

Thanks for review,
Sergo.


src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.h
```diff
@@ -0,0 +1,134 @@
...
+// THE SOFTWARE.
+//
+
+#pragma once
+
+// appleseed.foundation headers.
+#include "foundation/platform/windows.h"    // include before 3ds Max headers
+
+// 3ds Max headers.
+#include <renderelements.h>
```
dictoon: Order alphabetically.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -37,6 +37,7 @@
 #include "appleseedrenderer/projectbuilder.h"
 #include "appleseedrenderer/renderercontroller.h"
 #include "appleseedrenderer/tilecallback.h"
+#include "appleseedrenderelement/appleseedrenderelement.h"
```
dictoon: Order alphabetically.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -125,6 +129,10 @@ void* AppleseedRenderer::GetInterface(ULONG id)
 
         return static_cast<IInteractiveRender*>(m_interactive_renderer);
     }
+    if (id == IRenderElementCompatible::IID)
+    {
+        return new AppleseedRECompatible();
```
dictoon: Unline `AppleseedInteractiveRender` above, here we are returning a new instance each time. Is this intended?

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -240,6 +248,8 @@ int AppleseedRenderer::Open(
     int                     default_light_count,
     RendProgressCallback*   progress_cb)
 {
+    BroadcastNotification(NOTIFY_PRE_RENDER, static_cast<void*>(&rend_params));
```
dictoon: The cast is redundant: in C++ all pointers can be implicitly converted to `void*`.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1385,6 +1389,41 @@ namespace
         }
         else
         {
+            asr::AOVContainer aovs;
+
+            IRenderElementMgr* re_manager = GetCOREInterface()->GetCurRenderElementMgr();
+            re_manager->SetDisplayElements(false);
+            for (size_t i = 0, e = re_manager->NumRenderElements(); i < e; ++i)
```
dictoon: `NumRenderElements()` returns `int`; use `int` to iterate (and then you can remove a cast below).

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1385,6 +1389,41 @@ namespace
...
+            {
+                const asr::AOVFactoryRegistrar factory_registrar;
```
dictoon: Move the creation of the factory registrar out of the loop: creating a factory registrar is an extremely expensive operation as it will scan search paths for plugins, load DLLs, etc. In fact we might want to create the registrar once when the plugin is loaded rather than everytime a project is built.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1385,6 +1389,41 @@ namespace
...
+
+                auto* render_element = re_manager->GetRenderElement(static_cast<int>(i));
+                render_element->SetEnabled(false);
+
+                if (render_element->ClassID() == AppleseedRenderElement::get_class_id())
+                {
+                    AppleseedRenderElement* element = static_cast<AppleseedRenderElement*>(render_element);
+                    int aov_index = 0;
+                    element->GetParamBlock(0)->GetValueByName(L"aov_index", 0, aov_index, FOREVER);
+                    if (aov_index > 0 && aov_index <= factories.size())
```
dictoon: Signed vs unsigned comparison here. I'm surprised you're not getting a warning. You will probably need to cast `factories.size()` to `int`.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1385,6 +1389,41 @@ namespace
...
+                    {
+                        asf::auto_release_ptr<asr::AOV> aov_entity = factories[aov_index - 1]->create(asr::ParamArray());
+                        if (aovs.get_by_name(aov_entity->get_name()) == nullptr)
+                        {
+                            PBBitmap* p_bitmap = nullptr;
+                            render_element->GetPBBitmap(p_bitmap);
+                            if (p_bitmap != nullptr)
```
dictoon: Add braces around multi-line `if` body.

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,495 @@
...
+//
+// AppleseedRenderElement class implementation.
+//
+
+namespace
+{
+    
+    std::vector<std::string> get_aov_names()
+    {
+        const asr::AOVFactoryRegistrar factory_registrar;
```
dictoon: Same problem here. I'm not sure in which order you'll see my comments, but as I explained in another one, this is an extremely expensive operation.

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,495 @@
...
+        auto factories = factory_registrar.get_factories();
+        auto aov_names = std::vector<std::string>();
```
dictoon: Better:

    std::vector<std::string> aov_names;

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,495 @@
...
+        for (size_t i = 0, e = factories.size(); i < e; ++i)
```
dictoon: Use range-based for loops whenever possible:

    for (const auto& factory : factories)
    {
        ...
    }

(You might need to adjust the type of `factory`.)

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,495 @@
...
+            aov_names.push_back(factories[i]->get_model_metadata().get("label"));
+
+        return aov_names;
+    }
+
+    void init_combo_box(IParamMap2* param_map)
+    {
+        HWND dlg_hwnd = 0;
```
dictoon: I think that ultimately `HWND` is a pointer so `nullptr` would be more appropriate.

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,495 @@
...
+  : m_pblock(nullptr)
+{
+    if (!loading)
+    {
+        g_appleseed_renderelement_classdesc.MakeAutoParamBlocks(this);
+        SetElementName(AppleseedRenderElementFriendlyClassName);
+    }
+}
+
+AppleseedRenderElement::~AppleseedRenderElement()
```
dictoon: Remove the destructor if it does nothing?

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,495 @@
...
+    const Interval&   changeInt,
+    RefTargetHandle   hTarget,
+    PartID&           partID,
+    RefMessage        message,
+    BOOL              propagate)
+{
+    return REF_SUCCEED;
+}
+
+RefTargetHandle	AppleseedRenderElement::Clone(RemapDir& remap)
```
dictoon: Remove tabulation.

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,495 @@
...
+}
+
+void AppleseedRenderElement::SetEnabled(BOOL on)
+{ 
+    m_pblock->SetValue(ParamIdEnableOn, 0, on);
+}
+
+BOOL AppleseedRenderElement::IsEnabled() const
+{
+    int	is_on;
```
dictoon: Remove tabulation.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
usakhelo: This class is only used to check element compatibility. But new instance each time is not needed, that's true. Also I don't know if it's get deleted anywhere later.

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
usakhelo: Seems that `FactoryArrayType` can't be used with range loops, it says `error C3312: no callable 'begin' function found for type 'renderer::AOVFactoryRegistrar::FactoryArrayType'`

Should I do `auto factories = asf::array_vector<std::vector<asr::IAOVFactory*>>(g_aov_factory_registrar.get_factories());` ?

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
usakhelo: I moved it to the same place as `AppleseedRenderElementClassDesc g_appleseed_renderelement_classdesc;`

usakhelo:
Hi @dictoon ,
I fixed your comments and also added a commit to fix the crash during the interactive rendering. Turned out during the interactive rendering RenderElementManager is not available, so I added a nullptr check. Note that this crash is not the same as the described in the issue #156.

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -76,21 +78,20 @@ namespace
     
     std::vector<std::string> get_aov_names()
     {
-        const asr::AOVFactoryRegistrar factory_registrar;
-        auto factories = factory_registrar.get_factories();
-        auto aov_names = std::vector<std::string>();
-        for (size_t i = 0, e = factories.size(); i < e; ++i)
-            aov_names.push_back(factories[i]->get_model_metadata().get("label"));
+        auto factories = asf::array_vector<std::vector<asr::IAOVFactory*>>(g_aov_factory_registrar.get_factories());
```
dictoon: > Seems that FactoryArrayType can't be used with range loops, it says error C3312: no callable 'begin' function found for type 'renderer::AOVFactoryRegistrar::FactoryArrayType'

Oh ok, the code was fine then, sorry.

> Should I do auto factories = asf::array_vector<std::vector<asr::IAOVFactory*>>(g_aov_factory_registrar.get_factories()); ?

No please, that's a lot worse with allocations, constructions, destructions, copying... :)

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.h
```diff
@@ -132,3 +137,4 @@ class AppleseedRenderElementClassDesc
 };
 
 extern AppleseedRenderElementClassDesc g_appleseed_renderelement_classdesc;
+extern renderer::AOVFactoryRegistrar g_aov_factory_registrar;
```
dictoon: Add newline character.

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,492 @@
...
+asr::AOVFactoryRegistrar g_aov_factory_registrar;
+
+
+//
+// AppleseedRenderElement class implementation.
+//
+
+namespace
+{
+
```
dictoon: Remove blank line.

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,492 @@
...
+    void init_combo_box(IParamMap2* param_map)
+    {
+        HWND dlg_hwnd = nullptr;
+        if (param_map != nullptr)
+            dlg_hwnd = param_map->GetHWnd();
+        if (dlg_hwnd == nullptr)
+            return;
+
+        SendMessage(GetDlgItem(dlg_hwnd, IDC_COMBO_AOV_NAME), CB_RESETCONTENT, 0, 0);
+        if (true)
```
dictoon: Remove test?

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,492 @@
...
+                SendMessage(GetDlgItem(dlg_hwnd, IDC_COMBO_AOV_NAME), CB_SETCURSEL, -1, 0);
+        }
+    }
+
+    class AppleseedRenderElementParamDlgProc
+      : public ParamMap2UserDlgProc
+    {
+      public:
+        void DeleteThis() override
+        {
```
dictoon: Are we sure not doing anything here is correct?

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
```diff
@@ -0,0 +1,492 @@
...
+                    GetDlgItem(dlg_hwnd, IDC_COMBO_AOV_NAME),
+                    CB_ADDSTRING,
+                    0,
+                    reinterpret_cast<LPARAM>(utf8_to_wide(name).c_str()));
+            }
+
+            int output_index = param_map->GetParamBlock()->GetInt(ParamIdAOVIndex, 0, FOREVER);
+            DbgAssert(output_index >= 1);
+            output_index -= 1;
+            if (output_index < aov_names.size())
```
dictoon: Signed/unsigned mismatch?

src/appleseed-max-impl/appleseedrenderelement/appleseedrenderelement.cpp
usakhelo: I thought it's fine since instance of this class is only created as a local object in anonymous namespace (line 176).

dictoon: You're right, it's not an actual bug, however it's better to make the code as agnostic as possible to specific usage patterns, hence I suggest implementing this method (just `delete this` I presume?).

usakhelo: I think usage pattern should be change in this case, otherwise I can't open rendering dialog box, it crashes the max.
I'll add `delete this` and assign dialog procedure in a way it is done in other parts of the plugin.

PR #165, opened 2018-04-16
Fix re-building of Ship configuration

usakhelo:
@dictoon I think this change fixes Ship configuration rebuild issue


dictoon:
Awesome, tested and it works. How did you find and fix the issue?

PR #168, opened 2018-04-20
Add render stamp support

usakhelo:
Hi,

I added render stamp checkbox and an editbox in System rollout. Also TileCallbackFactory was added to work correctly with appleseed after-stamp callbacks.

Thanks for review!
Sergo.


src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -684,6 +684,7 @@ namespace
         ICustEdit*              m_text_renderingthreads;
         ISpinnerControl*        m_spinner_renderingthreads;
         AppleseedRenderer*      m_renderer;
+        ICustEdit*              m_text_render_stamp;
```
dictoon: Keep data members ordered logically: `m_renderer` probably needs to stay at the end and `m_text_render_stamp` should stay close to the other settings.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -706,11 +707,15 @@ namespace
         {
             ReleaseISpinner(m_spinner_renderingthreads);
             ReleaseICustEdit(m_text_renderingthreads);
+            ReleaseICustEdit(m_text_render_stamp);
             m_rend_params->DeleteRollupPage(m_rollup);
         }
 
         virtual void init(HWND hwnd) override
         {
+            m_text_render_stamp = GetICustEdit(GetDlgItem(hwnd, IDC_TEXT_RENDER_STAMP));
```
dictoon: Same thing here, keep initialization order consistent with the settings order, data member orders, etc.

src/appleseed-max-impl/appleseedrenderer/tilecallback.cpp
```diff
@@ -264,3 +264,26 @@ void TileCallback::blit_tile(
...
+}
+
+void TileCallbackFactory::release()
+{
+    delete this;
+}
+
+asr::ITileCallback* TileCallbackFactory::create()
+{
+    return new TileCallback(m_bitmap, m_rendered_tile_count);
```
dictoon: That may be incorrect.

There was a reason why we were passing a single tile callback instead of a factory: It's likely that Max doesn't support updating the rendered bitmap or the window from multiple threads, which is exactly what's going to happen with this change.

I don't know if Max supports that. If it doesn't, then that tile callback factory should always return the same tile callback instance (and make sure it only gets destroyed once of course).

usakhelo: I think it's better to update bitmap from single thread.
But how returning single instance of TileCallback prevent appleseed from running it from multiple threads?

dictoon: You're of course right, that was stupid of me :) We would need to enqueue the requests and run them from the main thread... Which is exactly what the `SerialTileCallback` does.

It seems to me that maybe the right solution would be to continue using a single tilecallback (instead of a factory), and fix `MasterRenderer` in the sense that it should execute all pending callback updates after adding the render stamp.

usakhelo:
Hi @dictoon,
I've checked recent changes to appleseed and it seems they fixed the issue. So I'm rolling back all the changes to TileCallback.

usakhelo:
Also addressing the issues you've noted.
About saving `m_log_open_mode` and `m_log_material_editor_messages` - these settings are saved into system ini file which makes those settings global across all the scenes.

dictoon:
> I've checked recent changes to appleseed and it seems they fixed the issue. So I'm rolling back all the changes to TileCallback.

Good to hear that they fix the issue! Thanks.

> About saving m_log_open_mode and m_log_material_editor_messages - these settings are saved into system ini file which makes those settings global across all the scenes.

Ok, that's perfect, thanks for the clarification.

usakhelo:
I've noticed in corona renderer stamp format editbox is not disabled when stamp is off. It seams reasonable. I'll just leave this here as a reminder:
image

dictoon:
After further discussion, this appears safe to merge so here we go.

PR #180, opened 2018-05-03
Fix select all shortcut in render log window

usakhelo:
This change should fix Ctrl-A hotkey in render log window.


src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.cpp
```diff
@@ -95,10 +95,18 @@ namespace
           case WM_COMMAND:
             switch (LOWORD(wparam))
             {
-              case ID_ACCELERATOR_LOG_CTRL_A:
+              case IDC_EDIT_LOG:
```
dictoon: So there's no need anymore for the `SendMessage()` call?

usakhelo: Yes, no more manual selecting, because windows handles it on its own. It's max that blocks all those calls. There should also be built-in context menu, which also blocked by something, I just don't know how to fix it.

PR #181, opened 2018-05-03
Add Volume Material

usakhelo:
This PR adds Volume material to 2016-2018 max plugins.
I'll remove 2015 plugin in following PR.

### DIFF
--- a/src/appleseed-max-impl/appleseedvolumemtl/appleseedvolumemtl.cpp
+++ b/src/appleseed-max-impl/appleseedvolumemtl/appleseedvolumemtl.cpp
@@ -1,0 +1,593 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2016-2018 Francois Beaune, The appleseedhq Organization
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+// Interface header.
+#include "appleseedvolumemtl.h"
+
+// appleseed-max headers.
+#include "appleseedrenderer/appleseedrenderer.h"
+#include "appleseedvolumemtl/datachunks.h"
+#include "appleseedvolumemtl/resource.h"
+#include "main.h"
+#include "oslutils.h"
+#include "utilities.h"
+#include "version.h"
+
+// appleseed.renderer headers.
+#include "renderer/api/material.h"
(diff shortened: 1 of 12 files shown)


src/appleseed-max-impl/appleseedvolumemtl/appleseedvolumemtl.cpp
```diff
@@ -0,0 +1,593 @@
...
+    m_params_validity.SetEmpty();
+
+    g_appleseed_volumemtl_classdesc.MakeAutoParamBlocks(this);
+}
+
+BaseInterface* AppleseedVolumeMtl::GetInterface(Interface_ID id)
+{
+    return
+        id == IAppleseedMtl::interface_id()
+        ? static_cast<IAppleseedMtl*>(this)
```
dictoon: Indent once more:

    return
        id == IAppleseedMtl::interface_id()
            ? static_cast<IAppleseedMtl*>(this)
            : Mtl::GetInterface(id);

src/appleseed-max-impl/appleseedvolumemtl/appleseedvolumemtl.cpp
```diff
@@ -0,0 +1,593 @@
...
+
+float AppleseedVolumeMtl::GetShinStr(int mtlNum, BOOL backFace)
+{
+    return 0.0f;
+}
+
+float AppleseedVolumeMtl::GetXParency(int mtlNum, BOOL backFace)
+{
+    float multiplier;
+    m_pblock->GetValue(ParamIdVolumeScatteringMultiplier, GetCOREInterface()->GetTime(), multiplier, FOREVER);
```
dictoon: We can do better by taking into using the luminance of `scattering color * scattering multiplier` instead of just relying on the multiplier.

src/appleseed-max-impl/appleseedvolumemtl/appleseedvolumemtl.cpp
```diff
@@ -0,0 +1,593 @@
...
+
+Color AppleseedVolumeMtl::GetAmbient(int mtlNum, BOOL backFace)
+{
+    return Color(0.0f, 0.0f, 0.0f);
+}
+
+Color AppleseedVolumeMtl::GetDiffuse(int mtlNum, BOOL backFace)
+{
+    Color color;
+    m_pblock->GetValue(ParamIdVolumeScattering, GetCOREInterface()->GetTime(), color, FOREVER);
```
dictoon: We can also do better here by taking into account the multiplier. Color and multiplier always go together.

src/appleseed-max-impl/appleseedvolumemtl/appleseedvolumemtl.cpp
```diff
@@ -0,0 +1,593 @@
...
+
+    // Absorption.
+    Color absorption;
+    m_pblock->GetValue(ParamIdVolumeAbsorption, time, absorption, m_params_validity);
+    const auto absorption_color_name = std::string(name) + "_volume_absorption";
+    insert_color(assembly, absorption, absorption_color_name.c_str());
+    volume_params.insert("absorption", absorption_color_name);
+
+    // Absorption Multiplier.
+    float absorption_multiplier = 0.0f;
```
dictoon: Do we need to initialize these variables?

usakhelo: Actually since we are dropping max 2015 I can use betteer versions, like `GetColor`, `GetFloat`, etc.

src/appleseed-max-impl/appleseedvolumemtl/appleseedvolumemtl.cpp
```diff
@@ -5,7 +5,7 @@
 //
 // This software is released under the MIT license.
 //
-// Copyright (c) 2016-2018 Francois Beaune, The appleseedhq Organization
+// Copyright (c) 2016-2018 Sergo Pogosyan, The appleseedhq Organization
```
dictoon: The year needs to be fixed too. Just take the line I was suggesting :)

src/appleseed-max-impl/appleseedvolumemtl/appleseedvolumemtl.cpp
```diff
@@ -387,9 +387,9 @@ Color AppleseedVolumeMtl::GetAmbient(int mtlNum, BOOL backFace)
 
 Color AppleseedVolumeMtl::GetDiffuse(int mtlNum, BOOL backFace)
 {
-    Color color;
-    m_pblock->GetValue(ParamIdVolumeScattering, GetCOREInterface()->GetTime(), color, FOREVER);
-    return color;
+    Color color = m_pblock->GetColor(ParamIdVolumeScattering, GetCOREInterface()->GetTime(), FOREVER);
```
dictoon: Make all variables const.

src/appleseed-max-impl/appleseedvolumemtl/appleseedvolumemtl.cpp
usakhelo: Sorry! I'm too impatient to read the whole comment :)

Issue #175, opened 2018-04-30
Project export fails when materials use Max procedural textures


Mango-3:
No appleseed project file is exported when the scene contains materials with MAX procedural textures assigned even though the scene renders without issue inside MAX (see screenshot)
I'm not sure if this is a bug of the current exporter or a feature which is not implemented. In the test scene I assigned two different substance materials which generate the textures for the appleseed DisneyMtl internally. 
Screenshot attached:
![scene_export_fails_max_proc_textures](https://user-images.githubusercontent.com/22252558/39445414-a220b736-4cc3-11e8-9c06-a33971a98771.PNG)

dictoon:
@usakhelo Can we disable (gray out) exporting when Use Max Procedural Maps is checked?

usakhelo:
I think we can.

Mango-3:
A quick fix would be to change the text next to the checkbox from "use Max Procedural Maps (non-compatible mode)" to "use Max Procedural Maps (disables scene export)"

dictoon:
@Mango-3 Totally! I never liked that "non-compatible mode" mention, non-compatible with what? I much prefer "disables scene export". We should do both: the mention and actually graying out export options. 

Thnaks for your input!

Mango-3:
Nicely done Sergo! Much less confusing now.

PR #186, opened 2018-05-08
Disable project exporting options when Use Max Procedural Maps turned on (issue #175)

usakhelo:
Scene export ui elements are disabled when Use Max Procedurals mode is selected.

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
@@ -565,10 +565,10 @@ namespace
 
         virtual void init(HWND hwnd) override
         {
+            m_static_project_filepath = GetDlgItem(hwnd, IDC_STATIC_PROJECT_FILEPATH);
             m_radio_render_only = GetDlgItem(hwnd, IDC_RADIO_RENDER);
             m_radio_save_only = GetDlgItem(hwnd, IDC_RADIO_SAVEPROJECT);
             m_radio_save_and_render = GetDlgItem(hwnd, IDC_RADIO_SAVEPROJECT_AND_RENDER);
-            m_static_project_filepath = GetDlgItem(hwnd, IDC_STATIC_PROJECT_FILEPATH);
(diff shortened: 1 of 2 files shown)


src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -562,6 +565,9 @@ namespace
 
         virtual void init(HWND hwnd) override
         {
+            m_radio_render_only = GetDlgItem(hwnd, IDC_RADIO_RENDER);
```
dictoon: We should always strive to keep orderings consistent, everywhere. Here, ordering of member declarations and ordering of initializations should match.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -583,18 +589,22 @@ namespace
             m_spinner_scale_multiplier->SetResetValue(RendererSettings::defaults().m_scale_multiplier);
             m_spinner_scale_multiplier->SetValue(m_settings.m_scale_multiplier, FALSE);
 
-            enable_disable_controls();
+            enable_disable_controls(false);
         }
 
-        void enable_disable_controls()
+        void enable_disable_controls(bool use_max_procedural_maps)
```
dictoon: Make arg const.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -563,6 +566,9 @@ namespace
         virtual void init(HWND hwnd) override
         {
             m_static_project_filepath = GetDlgItem(hwnd, IDC_STATIC_PROJECT_FILEPATH);
```
dictoon: `m_static_project_filepath` and `m_text_project_filepath` should probably stay together.

usakhelo: I'll move it here and change declaration order too...

PR #187, opened 2018-05-08
Fix render element rollout title

usakhelo:
Fixes render element title. I was empty on max 2016 and contained garbage on 2017.


dictoon:
Thanks!

PR #189, opened 2018-05-14
Fix built-in Plastic material in OSL mode

usakhelo:
There's unreported bug in Plastic material. When rendering in normal mode Diffuse and Specular colors and Roughness parameter do not affect the material.


src/appleseed-max-impl/appleseedplasticmtl/appleseedplasticmtl.cpp
```diff
@@ -683,6 +683,9 @@ asf::auto_release_ptr<asr::Material> AppleseedPlasticMtl::create_osl_material(
 
     shader_group->add_shader("surface", "as_max_plastic_material", name, 
         asr::ParamArray()
+            .insert("DiffuseColor", fmt_osl_expr(to_color3f(m_diffuse)))
```
dictoon: We should follow some logical ordering here, probably the ordering of shader inputs.

PR #192, opened 2018-05-24
Implement renderer UI with ParamBlock

usakhelo:
@dictoon 
Please take a look on this implementation of renderer's UI. There's a paramblock that describes all the ui elements and each element have an accessor which save and loads things into RendererSettings class. Right now it's fully compatible with scenes saved with original plugin.
I simplified appleseedrendererparameterdlg.cpp file. I'm not sure if we need Impl class in this case.
Save and Load function have not been changed, so everything is saved and loaded with RendererSettings class.
I'll try to add new parameters like override material and see if it works properly with old scenes.
Thanks,
Sergo.

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
@@ -38,6 +38,8 @@
 #include "appleseedrenderer/projectbuilder.h"
 #include "appleseedrenderer/renderercontroller.h"
 #include "appleseedrenderer/tilecallback.h"
+#include "main.h"
+#include "resource.h"
 #include "utilities.h"
 #include "version.h"
 
@@ -57,6 +59,7 @@
 #include <assert1.h>
 #include <bitmap.h>
 #include <interactiverender.h>
+#include <iparamm2.h>
 #include <notify.h>
 #include <pbbitmap.h>
 #include <renderelements.h>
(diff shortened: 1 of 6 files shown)


src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -928,7 +459,49 @@ AppleseedRendererParamDlg::AppleseedRendererParamDlg(
...
+        0,
+        renderer->GetParamBlock(0),
+        rend_params,
+        g_module,
+        MAKEINTRESOURCE(IDD_FORMVIEW_RENDERERPARAMS_OUTPUT),
+        L"Output",
+        0,
+        new OutputParamMapDlgProc(renderer->GetParamBlock(0)));
+
+   m_pmap_image_sampling = CreateRParamMap2(
```
dictoon: Can you order these calls (as well as all related stuff such as member variable declarations and initializations, etc.) according to the order of the panels in the UI? I.e. image sampling stuff first, etc. It makes things easier to find.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -938,10 +511,26 @@ AppleseedRendererParamDlg::~AppleseedRendererParamDlg()
 
 void AppleseedRendererParamDlg::DeleteThis()
 {
+    if (m_pmap_output != nullptr)
```
dictoon: Probably a good idea to destroy them in inverse order of creation.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.h
```diff
@@ -57,4 +57,10 @@ class AppleseedRendererParamDlg
   private:
     struct Impl;
     Impl* impl;
+
+    AppleseedRenderer*  m_renderer;
```
dictoon: Move to `Impl`?

dictoon:
Thanks! It looks good. A bit simpler indeed despite the PR adding 240 lines of code :) I prefer your version so let's merge this!

usakhelo:
Question on RendererSettings class - I like the idea of the separate class for holding all the properties and how we can have different default properties for interactive and regular renderer. But given that all the properties are now accessible from renderer's pblock, we can remove Settings class. I will need to fix loading older scenes but it's probably solvable. Again I like separate store class. The only upside from removing is that I could remove those getter/setter methods.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -74,13 +77,527 @@ namespace
...
+        ParamIdLogMaterialRendering     = 18,
+        ParamIdUseMaxProcedurals        = 19,
+        ParamIdEnableLowPriority        = 20,
+        ParamIdEnableRenderStamp        = 21,
+        ParamIdRenderStampFormat        = 22,
+    };
+
+    MCHAR* GetString(int id)
+    {
+        static wchar_t buf[256];
```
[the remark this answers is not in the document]
dictoon: > I thought static makes it safe, no?

I'm not sure what you mean by "safe". `static` just means that a single buffer is allocated, instead of one per invocations of `GetString()`. Declaring `buf` as static is the same as moving its declaration outside of `GetString()`. What this means is that in the following code:

    const MCHAR* s0 = GetString(0);
    const MCHAR* s1 = GetString(1);

`s0` and `s1` are equal, they both point to `buf`; as such the second call to `GetString()` invalidates the value returned by the first call.

dictoon:
> I will need to fix loading older scenes but it's probably solvable.

I'm not sure to understand.

I'm assuming that _right now_ (with the PR as it is now), compatibility is 100% preserved.

What is the change that you're describing that would break compatibility? Removing the `RendererSettings` class?

> The only upside from removing is that I could remove those getter/setter methods.

You mean the two big `Set()` and `Get()` methods in `AppleseedRendererPBlockAccessor`?

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -74,13 +77,527 @@ namespace
...
+
+void AppleseedRendererPBlockAccessor::Get(
+    PB2Value&       v,
+    ReferenceMaker* owner,
+    ParamID         id,
+    int             tab_index,
+    TimeValue       t,
+    Interval        &valid)
+{
+    AppleseedRenderer* const renderer = dynamic_cast<AppleseedRenderer*>(owner);
```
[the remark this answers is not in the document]
usakhelo: Great tip, thanks.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
usakhelo: I think they are in order, aren't they?  
m_pmap_output
m_pmap_image_sampling
m_pmap_lighting
m_pmap_system

image

usakhelo:
Hi @dictoon,
I've fixed the comments and also made at attempt to replace string resources with constants. I used KeyValuePair, but I couldn't find kvpair.h in api headers. Can we use it in plugin?

About removing the RendererSettings class - yes, I think removing that class will break compatibility. I'm not sure if we want to remove it. It was just a thought. And yes, Set() and Get() methods can be removed if we remove RendererSettings.

I'll add one more commit if everything else is fine.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -1307,5 +1313,11 @@ const MCHAR* AppleseedRendererClassDesc::InternalName()
 
 const MCHAR* AppleseedRendererClassDesc::GetRsrcString(INT_PTR id)
 {
-    return GetString(static_cast<int>(id));
+    const asf::KeyValuePair<int, const wchar_t*>* dialog_string_pair =
+        lookup_kvpair_array(
+            g_dialog_strings,
+            11,
```
usakhelo: This number should probably be replaced with constant. I'll fix it if everything else is fine.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.h
```diff
@@ -51,22 +51,49 @@
...
+        ParamID         id,
+        int             tab_index,
+        TimeValue       t) override;
+};
+
 class AppleseedRenderer
   : public Renderer
   , public ITabDialogObject
-{
+{
```
dictoon: Remove trailing spaces.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -264,310 +265,68 @@ namespace
...
-                    m_settings.m_max_ray_intensity = m_spinner_max_ray_intensity->GetFVal();
-                    return TRUE;
-
-                  case IDC_SPINNER_BACKGROUND_ALPHA:
-                    m_settings.m_background_alpha = m_spinner_background_alpha->GetFVal();
+                  case IDC_RADIO_SAVEPROJECT_AND_RENDER:
+                    enable_disable_controls();
                     return TRUE;
-
+
```
dictoon: Remove trailing spaces.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -576,310 +335,89 @@ namespace
...
-        {
-            ReleaseISpinner(m_spinner_scale_multiplier);
-            ReleaseICustEdit(m_text_scale_multiplier);
-            ReleaseICustButton(m_button_browse);
-            ReleaseICustEdit(m_text_project_filepath);
-            m_rend_params->DeleteRollupPage(m_rollup);
-        }
 
-        void init(HWND hwnd) override
+        void init(HWND hwnd)
```
dictoon: No longer overriding a base class method?

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -576,310 +335,89 @@ namespace
...
-        void enable_disable_controls(const bool use_max_procedural_maps)
+        void enable_disable_controls()
         {
+            DbgAssert(m_pblock != nullptr);
+            
+            int temp_use_max_procedural_maps;
+            int output_mode;
+            m_pblock->GetValueByName(L"use_max_procedural_maps", 0, temp_use_max_procedural_maps, FOREVER);
+            m_pblock->GetValueByName(L"output_mode", 0, output_mode, FOREVER);
+            bool use_max_procedural_maps = temp_use_max_procedural_maps > 0;
```
dictoon: Make const.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -74,13 +77,533 @@ namespace
...
+        ParamIdOpenLogMode              = 17,
+        ParamIdLogMaterialRendering     = 18,
+        ParamIdUseMaxProcedurals        = 19,
+        ParamIdEnableLowPriority        = 20,
+        ParamIdEnableRenderStamp        = 21,
+        ParamIdRenderStampFormat        = 22,
+    };
+    
+    const asf::KeyValuePair<int, const wchar_t*> g_dialog_strings[] =
+    {
```
dictoon: Does that compile with VS 2012?

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -74,13 +77,533 @@ namespace
...
+
+      case ParamIdProjectPath:
+        v.s = settings.m_project_file_path;
+        break;
+
+      case ParamIdScaleMultiplier:
+        v.f = settings.m_scale_multiplier;
+        break;
+        
+    //
```
dictoon: Align vertically with `case` statements.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -103,6 +628,36 @@ const RendererSettings& AppleseedRenderer::get_renderer_settings()
...
+    }
+}
+
+
+void AppleseedRenderer::SetReference(int i, RefTargetHandle rtarg)
+{
+    switch (i)
+    {
+      case 0:
+        m_param_block = dynamic_cast<IParamBlock2*>(rtarg);
```
dictoon: This will have the effect of setting `m_param_block` to `nullptr` if `rtarg` is not a `IParamBlock2*`. Do you know if this can ever happen? If not, then I would suggest a `static_cast<>` instead.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
dictoon: You can use `LOOKUP_KVPAIR_ARRAY()` which will infer the number of values in `g_dialog_strings` automatically.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
usakhelo: Yes, surprisingly, it does.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
usakhelo: Yes, this class does not inherit PanelBase. Actualy the only PanelBase class is AboutPanel. 
Output and System rollouts just use extended ParamMap2UserDlgProc classes.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
dictoon: That's good news!

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -728,3 +1314,11 @@ const MCHAR* AppleseedRendererClassDesc::InternalName()
     // Parsable name used by MAXScript.
     return L"appleseedRenderer";
 }
+
+const MCHAR* AppleseedRendererClassDesc::GetRsrcString(INT_PTR id)
+{
+    const asf::KeyValuePair<int, const wchar_t*>* dialog_string_pair =
```
dictoon: This is where C++11's `auto` comes handy:

    const auto dialog_string_pair = LOOKUP_KVPAIR_ARRAY(g_dialog_strings, id);

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.h
```diff
@@ -57,4 +57,9 @@ class AppleseedRendererParamDlg
   private:
     struct Impl;
     Impl* impl;
+
+    IParamMap2*         m_pmap_output;
```
dictoon: Why aren't these in `Impl`?

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -435,72 +434,92 @@ namespace
...
+        if (m_pmap_lighting != nullptr)
+            DestroyRParamMap2(m_pmap_lighting);
+
+        if (m_pmap_image_sampling != nullptr)
+            DestroyRParamMap2(m_pmap_image_sampling);
+
+        if (m_pmap_output != nullptr)
+            DestroyRParamMap2(m_pmap_output);
+
+        m_pmap_system = nullptr;
```
dictoon: Setting these to `nullptr` is not useful since the object is destructed anyway. It's a detail, you can remove them in a subsequent PR.

PR #195, opened 2018-07-30
Fix Tile::set_pixel call and render stamp

usakhelo:
Hi,
This pr fixes Tile::set_pixel call and render stamp.

Thank,
Sergo.


dictoon:
Thank you!

PR #196, opened 2018-08-01
Fix max2018 2019

usakhelo:
1. Quick fix of the crash that happens when accessing string parameters in 3ds max 2018, 2019
2. Fix maxscript access of the renderer parameters
3. Fix crash when opening log window after rendering

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
@@ -207,10 +207,6 @@ void AppleseedRendererPBlockAccessor::Get(
         v.i = static_cast<int>(settings.m_output_mode);
         break;
 
-      case ParamIdProjectPath:
-        v.s = settings.m_project_file_path;
-        break;
-
       case ParamIdScaleMultiplier:
         v.f = settings.m_scale_multiplier;
         break;
@@ -297,10 +293,6 @@ void AppleseedRendererPBlockAccessor::Get(
             
       case ParamIdEnableRenderStamp:
         v.i = static_cast<int>(settings.m_enable_render_stamp);
-        break;
-
-      case ParamIdRenderStampFormat:
-        v.s = settings.m_render_stamp_format;
         break;
 
       case ParamIdLogMaterialRendering:
--- a/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
@@ -1377,5 +1377,8 @@ const MCHAR* AppleseedRendererClassDesc::GetRsrcString(INT_PTR id)
 {
     const auto* dialog_string_pair = LOOKUP_KVPAIR_ARRAY(g_dialog_strings, id);
 
-    return dialog_string_pair->m_value;
+    if (dialog_string_pair != nullptr)
+        return dialog_string_pair->m_value;
+    else
+        return nullptr;
 }
--- a/src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
@@ -410,6 +410,7 @@ namespace
                 {
                   case IDC_BUTTON_LOG:
                     m_renderer->show_last_session_log();
+                    break;
 
                   case IDC_CHECK_USE_MAX_PROCEDURAL_MAPS:
                     {
--- a/src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
@@ -126,7 +126,7 @@ void RendererSettings::apply_settings_to_final_config(asr::Project& project) con
     asr::ParamArray& params = project.configurations().get_by_name("final")->get_parameters();
 
     params.insert_path("generic_frame_renderer.tile_ordering", "spiral");
-    params.insert_path("generic_frame_renderer.passes", m_passes);
+    params.insert_path("passes", m_passes);
     params.insert_path("shading_result_framebuffer", m_passes == 1 ? "ephemeral" : "permanent");
 
     params.insert_path("uniform_pixel_renderer.samples", m_pixel_samples);


src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -207,10 +207,6 @@ void AppleseedRendererPBlockAccessor::Get(
         v.i = static_cast<int>(settings.m_output_mode);
         break;
 
-      case ParamIdProjectPath:
```
dictoon: What are the consequences on functionality and backward compatibility of removing these two `case` blocks?

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -1377,5 +1377,8 @@ const MCHAR* AppleseedRendererClassDesc::GetRsrcString(INT_PTR id)
 {
     const auto* dialog_string_pair = LOOKUP_KVPAIR_ARRAY(g_dialog_strings, id);
 
-    return dialog_string_pair->m_value;
+    if (dialog_string_pair != nullptr)
```
dictoon: In what cases would we end up with a null pointer here?

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
```diff
@@ -410,6 +410,7 @@ namespace
                 {
                   case IDC_BUTTON_LOG:
                     m_renderer->show_last_session_log();
+                    break;
```
dictoon: Good catch!

There's a typo in the commit title (*openning* -> *opening*).

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
usakhelo: In cases of paramblock parameters when they don't have resource string ID.
In renderer pblock all the parameters have 0 for string id.

usakhelo: I couldn't find any use of sring resources for pblock params, so I decided to put 0 and it looks fine. Less resource entries looked good thing to me.

src/appleseed-max-impl/appleseedrenderer/appleseedrendererparamdlg.cpp
usakhelo: fixing it and force-pushing new commit.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
dictoon: This question still holds :)

usakhelo: Sorry I missed this one. Let me check, it didn't change the functionality, but I didn't try backward compatibility test.

usakhelo: @dictoon, I checked and it works fine. It opens the file from master version without any problem, settings are all there and everything.
It's still not clear why that section started to crash in 2018-19, but it looks like a fix, I don't know.

dictoon: I just don't understand how the project path is saved if `ParamIdProjectPath` is no longer handled. It looks like the project path should no longer be written into project files?

usakhelo: Hi @dictoon 
Setting project path variable happens in `AppleseedRendererPBlockAccessor::Set` function, and it has not been changed. So saving settings should be ok.
I thought loading may brake but it's ok too, so I think this change to Get methods seems harmless.

dictoon: From what I understand, `AppleseedRendererPBlockAccessor::Get()` extracts values from `owner` (of type `ReferenceMaker*`) and stores them in `v` (of type `PB2Value&`), so it seems that this corresponds to _saving_ settings. Conversely, `Set()` seems to correspond to _loading_ settings.

So as far as I can tell your change here should impact _saving_ project files, not loading them.

I still don't understand why everything would continue working with these changes (i.e. with hadling of `ParamIdProjectPath` removed). If it's indeed the case that things continue to work 100%, then isn't there a lot of other redundant code that could also be removed?

What's the logic?

usakhelo: This is how I understand it:
Saving and loading of renderer parameters is still happening in the `RendererSettings::save/load` functions, which saves, let's say `m_project_file_path` variable. Value of that variable is set in the `PBlockAccessor::Set()` function. When I enter the path into UI and hit enter `Set()` function is called and sets the value from `v.s` to `m_project_file_path`. After that our old saving part should correctly save the path.

Other thing is that these new pblock properties that I added recently are also saving/loading by themself. But I added `P_TRANSIENT` keyword to all of them, which according to docs stops them from being stored. Here's the quote "do not store actual value, PBAccessor-derived."

Also I added `P_CALLSETS_ON_LOAD` flag to the pblock header, which after loading the file should call `Set()` function to all the pblock properties.

I agree that it's not clear why everything works after removing two property getters. My guess is that properties are still saved on disk and get loaded, and that happens before `Get()/Set()` functions are called, so they only update the values with what our Save/Load functions get and put into renderersettings class.

usakhelo: About redundent code - we could probably remove all our `RendererSettings::Save/Load` functions and access all the properties directly from Renderer object but that would break backward compatibility.

dictoon: Thanks for the detailed explanation, much appreciated.

What I now understand better is that this code has nothing to do with loading/saving settings, instead it appears to be about feeding the paramblocks behind the user interface widgets.

What I still don't understand is why we can safely remove the code that handles project file paths but not code related to other settings.

Let's merge this. I'd suggest to keep investigating whether more code can be removed or not.

usakhelo: Crap, I forgot to merge this PR. Wandering what happens with Mango3's pr if I merge this now...

PR #200, opened 2018-08-16
Enable render controls for Embree, ray type bounce limits, direct & env light samples multiplier, roughness clamping, Russian Roulette start bounce

Mango-3:
Works on Max 2019, needs some checks and some code formatting


src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -69,6 +69,7 @@ namespace
             m_scale_multiplier = 1.0f;
 
             m_rendering_threads = 0;    // 0 = as many as there are logical cores
+			m_use_embree = true;
```
dictoon: We want Embree to be _disabled_ by default as it's still quite experimental and doesn't provide much performance benefits yet.

We use 4 spaces for indentation instead of tabs in all appleseed code bases. Replace tabs by spaces everywhere.

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -110,7 +111,7 @@ void RendererSettings::apply_common_settings(asr::Project& project, const char*
 
     params.insert_path("pt.enable_ibl", m_background_emits_light);
     params.insert_path("pt.enable_caustics", m_caustics);
-
+	params.insert_path("use_embree", m_use_embree);
```
dictoon: That line looks good but it's in the middle of the PT settings. Move it somewhere else, maybe right before or after setting `sampling_mode`?

src/appleseed-max/plugin.cpp
```diff
@@ -135,6 +135,8 @@ extern "C"
         static const wchar_t PluginImplDLLFilename[] = L"appleseed-max2017-impl.dll";
 #elif MAX_RELEASE == MAX_RELEASE_R20
         static const wchar_t PluginImplDLLFilename[] = L"appleseed-max2018-impl.dll";
+#elif MAX_RELEASE == MAX_RELEASE_R21
```
dictoon: Not related to Embree so I'd say this is in the wrong commit, but we can live with that. However are there not other changes required to support Max 2019? At least we need to provide a new VS solution... All this work should probably be done in another commit or even PR.

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -64,7 +64,8 @@ namespace
             m_background_emits_light = true;
             m_background_alpha = 0.0f;
             m_force_off_default_lights = false;
-
+			m_clamp_roughness = true;
```
dictoon: As this is a hack, it should be off by default (just like Max Ray Intensity).

dictoon: Place it next to Max Ray Intensity as they belong to the same category (performance hacks).

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -108,6 +109,7 @@ void RendererSettings::apply_common_settings(asr::Project& project, const char*
         params.insert_path("pt.max_diffuse_bounces", 0);
 
     params.insert_path("pt.max_bounces", m_bounces);
+	params.insert_path("pt.clamp_roughness", m_clamp_roughness);
```
dictoon: Move it next to MRI.

src/appleseed-max-impl/appleseedrenderer/renderersettings.h
```diff
@@ -78,7 +78,8 @@ class RendererSettings
     bool        m_background_emits_light;
     float       m_background_alpha;
     bool        m_force_off_default_lights;
-
+	bool		m_clamp_roughness;
```
dictoon: Move next to MRI.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -279,6 +277,10 @@ void AppleseedRendererPBlockAccessor::Get(
         v.f = settings.m_background_alpha;
         break;
 
+	  case ParamIdRoughnessClamping:
```
dictoon: Move next to MRI.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -404,6 +406,9 @@ void AppleseedRendererPBlockAccessor::Set(
         settings.m_background_alpha = v.f;
         break;
 
+	  case ParamIdRoughnessClamping:
+		  settings.m_clamp_roughness = v.i > 0;
+		  break;
```
dictoon: Keep the blank line before the block comment below.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -579,6 +588,12 @@ ParamBlockDesc2 g_param_block_desc(
         p_accessor, &g_pblock_accessor,
     p_end,
 
+	ParamIdRoughnessClamping, L"roughness_clamping", TYPE_BOOL, P_TRANSIENT, 0,
+		p_ui, ParamMapIdLighting, TYPE_SINGLECHEKBOX, IDC_CHECK_ROUGHNESS,
+		p_default, TRUE,
```
dictoon: Default should be FALSE.

src/appleseed-max-impl/appleseedrenderer/resource.h
```diff
@@ -2,80 +2,78 @@
...
-#define IDC_COMBO_LOG                               601
-#define IDC_STATIC_LOG                              602
-#define IDC_BUTTON_LOG                              603
-#define IDC_EDIT_LOG                                604
-#define IDC_CHECK_RENDER_STAMP                      605
-#define IDC_TEXT_RENDER_STAMP                       606
-#define IDS_RENDERERPARAMS_LOG_OPEN_MODE_1          607
-#define IDS_RENDERERPARAMS_LOG_OPEN_MODE_2          608
-#define IDS_RENDERERPARAMS_LOG_OPEN_MODE_3          609
+#define IDD_FORMVIEW_RENDERERPARAMS_ABOUT 100
```
dictoon: Formatting lost in this file...

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
Mango-3: Agree, I have set Embree to disabled by default.

src/appleseed-max/plugin.cpp
Mango-3: I can only test Max 2019, so I need to have this line to compile it. We can just comment it out on master for the time being. I have a VS solution of course, copied the one for Max 2018 and modified the link to the SDK to point to 2019 version. I do the development on a local copy of the project folder which gets pushed to GitHub and copy just the source files I had changed to it. This way I wanted to avoid that visual studio project files with project specific settings to my personal builds gets pushed to GitHub. I'm not sure this is the best way to do it. I haven't worked on a GitHub project before.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: I have set to default false in the new version

src/appleseed-max/plugin.cpp
usakhelo: @dictoon max2018 and 2019 have compatible sdk-s. But we still need a separate solution, right? To copy things correctly and everything..

src/appleseed-max-impl/appleseedrenderer/resource.h
Mango-3: I changed the formatting as you asked (replaced all tabs with spaces, regrouped  etc.) in the latest 2 commits of this PR.

src/appleseed-max/plugin.cpp
dictoon: Yes we need a separate solution.

PR #206, opened 2018-09-03
Minor code cleanup

Mango-3:


src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -92,13 +92,11 @@ namespace
         ParamIdOuputMode                    = 0,
         ParamIdProjectPath                  = 1,
         ParamIdScaleMultiplier              = 2,
-
```
dictoon: If you think these make sense (to separate different sections) you can keep them. I was more talking about the lonely blank line at the end of the enum. Not a big deal either way.

Mango-3: I will restore the section spacing, makes it more readable and remove the blank line with a new commit.

Issue #86, opened 2017-12-01
Add Preview Quality multiplier in settings


juancarlosgzrz:

juancarlosgzrz:
For material previews,

PR #213, opened 2018-09-21
Material preview render improvements

Mango-3:
-Add a material preview quality control (Issue #86  )
-Update material preview render defaults to safe settings
 (use always uniform sampling, disable shading overrides)


src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -1676,11 +1692,20 @@ int AppleseedRenderer::Render(
     // Retrieve and tweak renderer settings.
     RendererSettings renderer_settings = m_settings;
     if (m_rend_params.inMtlEdit)
-    {
-        renderer_settings.m_uniform_pixel_samples = m_rend_params.mtlEditAA ? 32 : 4;
+    {   
+        renderer_settings.m_lighting_algorithm = 0;
```
dictoon: May I suggest introducing constants (via an enum) for lighting algorithm, sampler type, shader override and denoise mode, and using them here?

PR #214, opened 2018-09-22
Improve and refactor packaging scripts

dictoon:


Mango-3:
Tested and works well with Max2019 and Max2018 builds on my system.

PR #216, opened 2018-09-22
Correct film dimensions for physical camera without DOF

Mango-3:


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1564,6 +1564,19 @@ void set_camera_dof_params(
     }
 }
 
+void set_camera_film_params(
```
dictoon: Nice! I would suggest changing `set_camera_dof_params()` to no longer set film params, and calling this function instead.

Mango-3:
Updated the set_camera_dof_params() function

PR #220, opened 2018-09-22
Refactorings and cleanups

dictoon:


Mango-3:
In the appleseed-max2019-impl.vcxproj,  appleseed-max2018-impl.vcxproj and appleseed-max2017-impl.vcxproj files:
lines 133 & 192 it states:
\appleseed-deps\stage\vc142\openexr-release\include; 
it should be: 
\appleseed-deps\stage\vc14\openexr-release\include;

Mango-3:
Looks all good to me.

PR #222, opened 2018-09-23
Map physical camera with DOF disabled to as_pinhole_camera

Mango-3:
Corrects the previously wrong mapping to as_thinlens_camera when DOF is disabled.

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -1609,19 +1609,24 @@ asf::auto_release_ptr<asr::Camera> build_camera(
         {
             // Film dimensions.
             set_camera_film_params(params, phys_camera, bitmap, time);
-            // Create camera.
-            camera = asr::PinholeCameraFactory().create("camera", params);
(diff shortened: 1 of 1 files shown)


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1609,18 +1609,19 @@ asf::auto_release_ptr<asr::Camera> build_camera(
         {
             // Film dimensions.
             set_camera_film_params(params, phys_camera, bitmap, time);
-
+            // Create camera.
```
dictoon: Insert blank lines before comments.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1609,18 +1609,19 @@ asf::auto_release_ptr<asr::Camera> build_camera(
...
+            camera = asr::PinholeCameraFactory().create("camera", params);
```
dictoon: We should avoid creating a camera and then immediately replacing it by another.

dictoon:
Thanks!

PR #224, opened 2018-09-24
Fix camera scale multiplier, search path separator

Mango-3:
- Scale Multiplier is now honored when setting film dimensions
- Search path separators converted to UNIX style
- Provisional fix for the material preview rendering. It has no side effects on the main scene or scene export and thus is safe to use. It can also be reverted at a later point without braking the scene or settings when the original cause is discovered and remedied.

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -632,20 +632,8 @@ namespace
                 "medium_priority",
                 get_medium_priority(instance_node->GetObjectRef(), time));
         if (type == RenderType::MaterialPreview)
-        {
             params.insert_path("visibility.shadow", false);
-            // Create the instance and insert it into the assembly.
-            assembly.object_instances().insert(
-                asr::ObjectInstanceFactory::create(
-                    instance_name.c_str(),
-                    params,
-                    object_info.m_name.c_str(),
-                    transform,
-                    back_material_mappings,
-                    front_material_mappings));
-        }
-        else
-        {
+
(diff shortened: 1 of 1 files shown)


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1532,10 +1545,11 @@ void set_camera_film_params(
     asr::ParamArray&            params,
     MaxSDK::IPhysicalCamera*    camera_node,
     Bitmap*                     bitmap,
+    const RendererSettings& settings,
```
dictoon: Align argument.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -632,17 +632,30 @@ namespace
...
-        // Create the instance and insert it into the assembly.
-        assembly.object_instances().insert(
-            asr::ObjectInstanceFactory::create(
-                instance_name.c_str(),
-                params,
-                object_info.m_name.c_str(),
-                transform,
-                front_material_mappings,
-                back_material_mappings));
+            // Create the instance and insert it into the assembly.
```
dictoon: This is a workaround. We must understand what caused the problem in the first place, maybe there's a bug somewhere else and I would rather fix the original problem than hack around it.

That said, the rest of the PR looks good so I would suggest to remove this part and keep (and merge) the rest.

Thanks!

dictoon: Revert indentation.

PR #231, opened 2018-10-01
Small Fixes

Mango-3:
- Correct Filter Size lower limit
- Rename Max Global Bounces to Max Ray Bounces


src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -1006,7 +1006,7 @@ ParamBlockDesc2 g_param_block_desc(
         p_accessor, &g_pblock_accessor,
     p_end,
 
-    ParamIdGlobalBounceLimit, L"global_illumination_bounces", TYPE_INT, P_TRANSIENT, 0,
+    ParamIdGlobalBounceLimit, L"max_ray_bounces", TYPE_INT, P_TRANSIENT, 0,
```
dictoon: Are we breaking any form of backward compatibility with this change?

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.rc
```diff
@@ -116,7 +116,7 @@ IDD_FORMVIEW_RENDERERPARAMS_PATH_TRACING DIALOGEX 0, 0, 200, 210
 STYLE DS_SETFONT | WS_CHILD | WS_VISIBLE
 FONT 8, "MS Sans Serif", 0, 0, 0x1
 BEGIN
-    LTEXT           "Max Global Bounces:",IDC_STATIC_BOUNCES,0,35,68,8
+    LTEXT           "Max Ray Bounces:",IDC_STATIC_BOUNCES,0,35,61,8
     CONTROL         "Max Global Bounces",IDC_TEXT_BOUNCES,"CustEdit",WS_TABSTOP,105,34,16,10
```
dictoon: Rename other associated controls.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: I haven't had issues loading older scenes but I'm not 100% sure. I will revert it back just to be on the safe side.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.rc
Mango-3: ok

dictoon:
Thanks! Could you combine the last two commits please?

Issue #212, opened 2018-09-21
Animated material parameters don't work


dictoon:
From @frang on Discord:

> I've been testing the appleseed-max2017-1.0.0-beta-preview1. I noticed something that doesn't work: when you animate a material parameter and you render a series of frames, the animated param doesn't change along them. If you render frames individually it works.

usakhelo:
I confirm this and this is applied to OSL materials.

dictoon:
Done.

PR #234, opened 2018-10-14
Update animated parameters by getting current time from renderer

usakhelo:
Hi,
This PR fixes bug #212. OSL materials and textures were not updated during animation rendering. Now they are working. What I did is instead of getting current with the call `GetCOREInterface()->GetTime()` I added a function to renderer to get currently rendered frame from the it.

Thanks for review!
Sergo.

### DIFF
--- a/src/appleseed-max-impl/appleseedenvmap/appleseedenvmap.cpp
+++ b/src/appleseed-max-impl/appleseedenvmap/appleseedenvmap.cpp
@@ -550,11 +550,10 @@ Point3 AppleseedEnvMap::EvalNormalPerturb(ShadeContext& /*sc*/)
     return Point3(0.0f, 0.0f, 0.0f);
 }
 
-asf::auto_release_ptr<asr::EnvironmentEDF> AppleseedEnvMap::create_envmap(const char* name)
+asf::auto_release_ptr<asr::EnvironmentEDF> AppleseedEnvMap::create_envmap(const char* name, const TimeValue time)
 {
     float sun_theta_deg = m_sun_theta;
     float sun_phi_deg = m_sun_phi;
-    const TimeValue time = get_current_time();
 
     if (m_sun_node != nullptr && m_sun_node_on)
     {
--- a/src/appleseed-max-impl/appleseedenvmap/appleseedenvmap.cpp
+++ b/src/appleseed-max-impl/appleseedenvmap/appleseedenvmap.cpp
@@ -554,10 +554,11 @@ asf::auto_release_ptr<asr::EnvironmentEDF> AppleseedEnvMap::create_envmap(const 
 {
     float sun_theta_deg = m_sun_theta;
     float sun_phi_deg = m_sun_phi;
+    const TimeValue time = get_current_time();
 
     if (m_sun_node != nullptr && m_sun_node_on)
     {
-        Matrix3 sun_transform = m_sun_node->GetObjTMAfterWSM(GetCOREInterface()->GetTime());
+        Matrix3 sun_transform = m_sun_node->GetObjTMAfterWSM(time);
         sun_transform.NoTrans();
 
         const Point3 sun_dir = (Point3::ZAxis * sun_transform).Normalize();
--- a/src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
+++ b/src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
@@ -87,8 +87,8 @@ OSLMaterial::OSLMaterial(Class_ID class_id, OSLPluginClassDesc* class_desc)
     m_has_bump_params = m_shader_info->find_maya_attribute("normalCamera") != nullptr;
     m_has_normal_params = m_shader_info->find_param("Tn") != nullptr;
 
+    m_params_validity.SetEmpty();
     class_desc->MakeAutoParamBlocks(this);
-    Reset();
 }
 
 BaseInterface* OSLMaterial::GetInterface(Interface_ID id)
@@ -202,10 +202,11 @@ RefResult OSLMaterial::NotifyRefChanged(
         break;
 
       case REFMSG_CHANGE:
-          if (hTarget == m_pblock)
-              m_class_desc->GetParamBlockDesc(0)->InvalidateUI(m_pblock->LastNotifyParamID());
-          else if (hTarget == m_bump_pblock)
-              m_class_desc->GetParamBlockDesc(1)->InvalidateUI(m_bump_pblock->LastNotifyParamID());
+        m_params_validity.SetEmpty();
+        if (hTarget == m_pblock)
+            m_class_desc->GetParamBlockDesc(0)->InvalidateUI(m_pblock->LastNotifyParamID());
+        else if (hTarget == m_bump_pblock)
+            m_class_desc->GetParamBlockDesc(1)->InvalidateUI(m_bump_pblock->LastNotifyParamID());
         break;
     }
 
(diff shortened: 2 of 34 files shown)


src/appleseed-max-impl/appleseedenvmap/appleseedenvmap.cpp
```diff
@@ -554,10 +554,11 @@ asf::auto_release_ptr<asr::EnvironmentEDF> AppleseedEnvMap::create_envmap(const
 {
     float sun_theta_deg = m_sun_theta;
     float sun_phi_deg = m_sun_phi;
+    const TimeValue time = get_current_time();
```
dictoon: Move inside the `if` below?

src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
```diff
@@ -279,30 +280,31 @@ TSTR OSLMaterial::GetSubTexmapSlotName(int i)
 
 void OSLMaterial::Update(TimeValue t, Interval& valid)
 {
+
```
dictoon: Remove redundant blank line.

src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
```diff
@@ -487,26 +489,26 @@ asf::auto_release_ptr<asr::Material> OSLMaterial::create_osl_material(
     // Shader group.
     //
 
-    const auto t = GetCOREInterface()->GetTime();
-
     auto shader_group_name = make_unique_name(assembly.shader_groups(), std::string(name) + "_shader_group");
     auto shader_group = asr::ShaderGroupFactory::create(shader_group_name.c_str());
+    
+    const TimeValue time = get_current_time();
```
dictoon: Move inside `if` below?

src/appleseed-max-impl/appleseedoslplugin/oslmaterial.cpp
```diff
@@ -279,30 +280,31 @@ TSTR OSLMaterial::GetSubTexmapSlotName(int i)
...
-            tex_map->Update(t, valid);
+        for (const auto& tex_param : m_texture_id_map)
+        {
+            Texmap* tex_map = nullptr;
+            if (m_has_bump_params && tex_param == m_texture_id_map.back())
+                m_bump_pblock->GetValue(tex_param.first, t, tex_map, m_params_validity);
+            else
+                m_pblock->GetValue(tex_param.first, t, tex_map, m_params_validity);
+
+            if (tex_map)
```
dictoon: Make comparison explicit.

src/appleseed-max-impl/appleseedoslplugin/osltexture.cpp
```diff
@@ -284,34 +285,35 @@ void OSLTexture::Update(TimeValue t, Interval& valid)
...
-    {
-        Texmap* tex_map = nullptr;
-        m_pblock->GetValue(tex_param.first, t, tex_map, valid);
-        if (tex_map)
-            tex_map->Update(t, valid);
+        for (auto& tex_param : m_texture_id_map)
+        {
+            Texmap* tex_map = nullptr;
+            m_pblock->GetValue(tex_param.first, t, tex_map, m_params_validity);
+            if (tex_map)
```
dictoon: Make comparison explicit.

src/appleseed-max-impl/appleseedoslplugin/osltexture.cpp
```diff
@@ -418,6 +420,8 @@ void OSLTexture::create_osl_texture(
     const char*         material_input_name,
     const int           output_slot_index)
 {
+    const TimeValue time = get_current_time();
```
dictoon: Remove, unused.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.h
```diff
@@ -172,6 +172,7 @@ class AppleseedRenderer
 
     void show_last_session_log();
     void create_log_window();
+    TimeValue get_time();
```
dictoon: Move somewhere more logical. Maybe after `get_renderer_settings()`?

Also, make const.

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -50,17 +50,17 @@ void connect_output_selector(
     const char*         material_input_name,
     Texmap*             texmap)
 {
-    const auto t = GetCOREInterface()->GetTime();
+    TimeValue time = get_current_time();
```
dictoon: Make const.

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -106,11 +107,11 @@ void connect_output_map(
     Texmap*             texmap,
     const float         const_value)
 {
-    const auto t = GetCOREInterface()->GetTime();
+    const auto time = get_current_time();
```
dictoon: Use `TimeValue` out of consistency with the rest of the code.

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -81,11 +81,12 @@ void connect_output_map(
     Texmap*             texmap,
     const Color         const_value)
 {
-    const auto t = GetCOREInterface()->GetTime();
+    TimeValue time = get_current_time();
```
dictoon: Make const.

src/appleseed-max-impl/osloutputselectormap/osloutputselector.cpp
```diff
@@ -438,22 +438,21 @@ void OSLOutputSelector::Update(TimeValue t, Interval& valid)
...
-
-    Texmap* source_map;
-    m_pblock->GetValue(ParamIdSourceMap, t, source_map, m_params_validity);
 
-    if (source_map)
-        source_map->Update(t, m_params_validity);
+        Texmap* source_map;
+        m_pblock->GetValue(ParamIdSourceMap, t, source_map, m_params_validity);
 
+        if (source_map)
```
dictoon: Make comparison explicit.

src/appleseed-max-impl/utilities.cpp
```diff
@@ -31,6 +31,7 @@
 
 // appleseed-max headers.
 #include "appleseedoslplugin/osltexture.h"
+#include "appleseedrenderer/appleseedrenderer.h"
```
dictoon: I find it a bit concerning that `utilities` depends on `appleseedrenderer`: lower level components should never depend on higher level ones. Maybe `get_current_time()` should be declared in `appleseedrenderer.{h,cpp}`.

For the same reason I'm slightly concerned about the dependencies toward `appleseedoslplugin/osltexture.h` and `osloutputselectormap/osloutputselector.h`.

usakhelo: Can I make `get_current_time` in appleseedrenderer a global function, not the class member? So that I don't have to check and cast current renderer to get the time.

dictoon: Yes, that's what I meant. You could also make it a static method of `AppleseedRenderer` to highlight the dependency to that class.

usakhelo: I'm trying to move it to the renderer class but in some utilities I still need to call that function so I still need to include renderer header.

usakhelo: Other way is to pass `time` from the renderer to all that `create_material` and so on as an argument.

usakhelo:
Hi @dictoon ,
I added time parameter to the utility functions and then I noticed that only couple of other functions need to be changed in order to get rid of static `get_time` function. So I decided to change all the functions and now `time` is passed from the project builder to each material creation function.
It may seem too much of parameters to some of the functions, but it also makes it obvious that they all need time information to work correctly.

luisbarrancos:
This might be relevant:
https://github.com/imageworks/OpenShadingLanguage/pull/828
OSL will have acess to global (or camera) frame.

dictoon:
> So I decided to change all the functions and now time is passed from the project builder to each material creation function.

In my opinion, definitely the right solution as it makes dependencies clear.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -1953,7 +1948,6 @@ void AppleseedRenderer::clear()
     m_entities.clear();
 }
 
-
```
dictoon: We conventionally keep two blank lines before block comments.

src/appleseed-max-impl/utilities.cpp
```diff
@@ -372,22 +373,14 @@ std::string insert_bitmap_texture_and_instance(
...
-}
-
 namespace
 {
     class MaxShadeContext
       : public ShadeContext
     {
       public:
-        explicit MaxShadeContext(const asr::SourceInputs& source_inputs)
+        explicit MaxShadeContext(const asr::SourceInputs& source_inputs, TimeValue time)
```
dictoon: `explicit` has become redundant here, not that it really hurts...

dictoon: Make `time` const.

src/appleseed-max-impl/utilities.cpp
```diff
@@ -567,9 +559,10 @@ namespace
       : public asr::Source
     {
       public:
-        explicit MaxProceduralTextureSource(Texmap* texmap)
+        explicit MaxProceduralTextureSource(Texmap* texmap, TimeValue time)
```
dictoon: Make `time` const.

src/appleseed-max-impl/utilities.cpp
```diff
@@ -651,18 +644,19 @@ namespace
         }
 
       private:
-        Texmap* m_texmap;
+        Texmap*     m_texmap;
+        TimeValue   m_time;
```
dictoon: Can it be made const here too? (I think so.)

src/appleseed-max-impl/utilities.cpp
```diff
@@ -689,9 +683,10 @@ namespace
       : public asr::Texture
     {
       public:
-        MaxProceduralTexture(const char* name, Texmap* texmap)
+        MaxProceduralTexture(const char* name, Texmap* texmap, TimeValue time)
```
dictoon: Make `time` const.

PR #250, opened 2018-11-21
Add On Warning option to Open Log menu

### LINKED ISSUE #190: Log window should not open in case of mere warnings, only errors

Mango-3:
- Added a dedicated `On Warning` menu entry to the `Open Log:` mode menu in the `System` panel.
Previously the log window opened and printed warnings when `On Error` only was selected.
This fixes #190


src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -188,6 +188,7 @@ namespace
         { IDS_RENDERERPARAMS_LOG_OPEN_MODE_1,       L"Always" },
         { IDS_RENDERERPARAMS_LOG_OPEN_MODE_2,       L"Never" },
         { IDS_RENDERERPARAMS_LOG_OPEN_MODE_3,       L"On Error" },
+        { IDS_RENDERERPARAMS_LOG_OPEN_MODE_4,       L"On Warning" },
```
dictoon: As suggested in the ticket, we should reorder items from most verbose to least verbose.

Also, I believe that reordering values in the `OpenMode` enumeration (while maintaing backward compatibility if possible) would allow to slightly simplify the logic in `DialogLogTarget::write()` since you could then write things like

    if (m_open_mode >= OpenMode::Errors)
        print_to_dialog();

Mango-3: I think that is not easily possible without braking compatibility with older scenes. 
The reordering of the enum `IDS_RENDERERPARAMS_LOG_OPEN_MODE_x` does not change the way the log-settings and the mapping to the log-levels are stored in the file.
One would need to change the ordering of the enum Class `DialogLogTarget::OpenMode` where there is the warning: // Changing these value WILL break compatibility.

dictoon: Ok, thanks. Can we at least reorder the items in the dropdown, without changing their values?

Mango-3: I tried reordering the dropdown list but it always ends up then being assigned to the wrong log mode. Maybe I missed something obvious how to do it easily. As long as the storage in the parameter block is not changed, old scenes should be compatible with it.

dictoon: Alright. Let's merge this as is since it looks good. Thanks!

PR #257, opened 2018-11-29
Expose noise seed and light sampler render controls

Mango-3:
- Added `Noise Seed` controls (setting a dedicated noise seed and enable variation per frame) 
- Added `Light Sampling` algorithm type menu
- Added checkbox to enable importance sampling for lights


src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -1178,6 +1187,13 @@ ParamBlockDesc2 g_param_block_desc(
         p_accessor, &g_pblock_accessor,
     p_end,
 
+    ParamIdNoiseSeed, L"noise_seed", TYPE_INT, P_TRANSIENT, 0,
+        p_ui, ParamMapIdImageSampling, TYPE_SPINNER, EDITTYPE_INT, IDC_TEXT_NOISE_SEED, IDC_SPINNER_NOISE_SEED, SPIN_AUTOSCALE,
+        p_default, 0,
+        p_range, 0, 10000000,
```
dictoon: If possible, the range should be [0, 4294967295], or if that's not possible because the underlying type is a 32-bit _signed_ integer, then it should be [-2147483648, 2147483647].

src/appleseed-max-impl/appleseedrenderer/resource.h
```diff
@@ -19,6 +19,7 @@
 #define IDC_SPINNER_FILTER_SIZE                         209
 #define IDS_RENDERERPARAMS_FILTER_TYPE_1                210
 #define IDS_RENDERERPARAMS_FILTER_TYPE_2                211
+#define IDC_SPINNER_FILTER_SIZE2                        211
```
dictoon: Remove?

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -2058,6 +2058,9 @@ int AppleseedRenderer::Render(
     if (!m_rend_params.inMtlEdit || m_settings.m_log_material_editor_messages)
         create_log_window();
 
+    if (renderer_settings.m_noise_seed == 0)
```
dictoon: As we discussed, noise seed = 0 shouldn't be a special value. Automatic variation of noise seed per frame should be controlled with a checkbox.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -2058,6 +2058,9 @@ int AppleseedRenderer::Render(
...
+        renderer_settings.m_noise_seed = static_cast<int>(time) / GetTicksPerFrame();
```
dictoon: Not sure the division makes sense. Noise seed should vary for every frame, not just every second.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -110,9 +110,12 @@ namespace
         ParamIdAdaptiveTileNoiseThreshold               = 44,
         ParamIdBackgroundAlphaValue                     = 15,
         ParamIdNoiseSeed                                = 76,
+        ParamIdEnableNoiseSeed                          = 77,
```
dictoon: That should be part of the previous commit.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -110,9 +110,12 @@ namespace
...
 
         ParamIdLightingAlgorithm                        = 52,
         ParamIdForceDefaultLightsOff                    = 13,
+        ParamIdLightSamplingAlgorithm                   = 78,
+        ParamIdEnableImportanceSampling                 = 79,
```
dictoon: We need to be more specific here: `ParamIdEnableLightImportanceSampling`.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -389,6 +398,14 @@ void AppleseedRendererPBlockAccessor::Get(
         v.i = static_cast<int>(settings.m_force_off_default_lights);
         break;
 
+      case ParamIdLightSamplingAlgorithm:
+        v.i = settings.m_light_sampling_algorithm;
+        break;
+
+      case ParamIdEnableImportanceSampling:
+        v.i = static_cast<int>(settings.m_enable_importance_sampling);
```
dictoon: Similarly, `settings.m_enable_light_importance_sampling`.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
dictoon: `ParamIdEnableNoiseSeed` is not quite accurate: it should be something like `ParamIdEnablePerFrameNoiseSeedVariation`.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.rc
```diff
@@ -114,6 +114,8 @@ BEGIN
     CONTROL         "Noise Seed",IDC_TEXT_NOISE_SEED,"CustEdit",WS_TABSTOP,50,162,30,10
     CONTROL         "Noise Seed Spinner",IDC_SPINNER_NOISE_SEED,
                     "SpinnerControl",WS_TABSTOP,82,162,6,10
+    CONTROL         "Vary Noise per Frame",IDC_CHECK_ENABLE_NOISE_SEED,
```
dictoon: Maybe "Vary Noise Pattern per Frame"?

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -184,6 +187,18 @@ const char* get_lighting_engine_type(const int lighting_engine_type)
     }
 }
 
+const char* get_lighting_algorithm_type(const int lighting_algorithm_type)
+{
+    switch (lighting_algorithm_type)
+    {
+    case 0:  return "cdf";
```
dictoon: Indent `case` and `default statements with 2 spaces. Check how other `switch` statements are indented.

src/appleseed-max-impl/appleseedrenderer/resource.h
```diff
@@ -43,6 +42,7 @@
 #define IDC_SPINNER_ADAPTIVE_NOISE_THRESHOLD            232
 #define IDC_TEXT_NOISE_SEED                             233
 #define IDC_SPINNER_NOISE_SEED                          234
+#define IDC_CHECK_ENABLE_NOISE_SEED                     235
```
dictoon: Why are some constants prefixed with `RENDERERPARAMS` while others not?

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: That is exactly what it does. In Max time is represented in ticks and a frame duration can vary in time depending on the frames/second setting (24,25 or 30). The GetTicksPerFrame() call returns the duration of the current frame in ticks.  By dividing the accumulated time (in ticks) by the frame duration (in ticks) I get the current frame number which varies the noise seed.
`renderer_settings.m_noise_seed += static_cast<int>(time) / GetTicksPerFrame();`

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: ok

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: ok

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: ok

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
dictoon: Got it, thanks for the explanation! I was thrown off by `GetTicksPerFrame()`, I somehow read it as `GetTicksPerSecond()` or something.

May I suggest introducing a temporary variable `frame_number` to clarify that logic?

src/appleseed-max-impl/appleseedrenderer/resource.h
Mango-3: RENDERPARAMS prefixed ones refer to the ui-element resources which have to be defined as key-value pairs for combo-box selection menus in appleseedrenderer.cpp.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: ok, I added a temporary variable.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.rc
dictoon: Ok.

src/appleseed-max-impl/appleseedrenderer/resource.h
dictoon: Ok.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -109,13 +109,13 @@ namespace
         ParamIdAdaptiveTileMaxSamples                   = 43,
         ParamIdAdaptiveTileNoiseThreshold               = 44,
         ParamIdBackgroundAlphaValue                     = 15,
-        ParamIdNoiseSeed                                = 76,
-        ParamIdEnableNoiseSeed                          = 77,
+        ParamIdPerFrameNoiseSeedVariation               = 76,
```
dictoon: The name for this one was correct before. It's not per-frame noise seed variation but the opposite: a fixed noise seed for all frames.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -1216,14 +1216,14 @@ ParamBlockDesc2 g_param_block_desc(
         p_accessor, &g_pblock_accessor,
     p_end,
 
-    ParamIdNoiseSeed, L"noise_seed", TYPE_INT, P_TRANSIENT, 0,
+    ParamIdPerFrameNoiseSeedVariation, L"noise_seed", TYPE_INT, P_TRANSIENT, 0,
         p_ui, ParamMapIdImageSampling, TYPE_SPINNER, EDITTYPE_INT, IDC_TEXT_NOISE_SEED, IDC_SPINNER_NOISE_SEED, SPIN_AUTOSCALE,
         p_default, 0,
-        p_range, 0, 2000000000,
+        p_range, 0, 2147483647,
```
dictoon: Minimum value should be -2147483648 if possible.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -1251,7 +1251,7 @@ ParamBlockDesc2 g_param_block_desc(
         p_accessor, &g_pblock_accessor,
     p_end,
 
-    ParamIdEnableImportanceSampling, L"enable_importance_sampling", TYPE_BOOL, P_TRANSIENT, 0,
+    ParamIdEnableLightImportanceSampling, L"enable_importance_sampling", TYPE_BOOL, P_TRANSIENT, 0,
```
dictoon: `L"enable_importance_sampling"` needs fixing too.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: True, copied and pasted it before too fast. I reverted it.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: changed it

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: I leave the lower value 0. Visual Studio complained when I set the lower value to -2,147,483,64. The range from 0 on upwards is really more than enough for any realistic animation ever to be rendered with appleseed-Max. Max does not recognize uint types in some ui-parameter block settings like p_range, only integer type but the render setting parameter it refers to can be both int or uint.

dictoon: It's not so much about the number of possible values, obviously 2 billion values is more than enough. It's about being able to match any valid noise seed value, and about not introducing artificial limits.

You indeed cannot use -2147483648 directly in C/C++, even with `L` suffix: `-2147483648L`. You would need two L suffixes: `-2147483648LL`, but that would make the constant a `long long` (64-bit).

The cleanest option is to use `numeric_limits<int>::min()` and `numeric_limits<int>::max()` (you'll need to include `<limits>`).

If you look at how they are defined, they use C's `INT_MIN` and `INT_MAX` macros, where `INT_MIN` is typically defined as `(-2147483647 - 1)`.

dictoon: I was curious for why one can't directly write -2147483648 in C, here are some answers:

http://brnz.org/hbr/?p=1451
https://stackoverflow.com/a/9941519/393756

dictoon:
The PR should be rebased to eliminate the last commit and to make all commits self-contained.

Thanks for the top-quality work!

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: Interesting, the reason why VC++ has INT_MIN defined that way. 
But I keep the lower bound to 0. I just checked why negative values don't work and saw that the noise seed parameter has been defined in appleseed core as a 32bit unsigned int in `src/appleseed/renderer/modeling/frame/frame.cpp` line 932 for instance:
`impl->m_noise_seed = m_params.get_optional<uint32>("noise_seed", 0);`

dictoon: In appleseed core it's an unsigned 32-bit value. If you set the noise seed to a negative value, it will simply wrap around. All 32 bits are preserved.

With the range [0, 2147483647] you effectively limit the choice to a 31-bit value.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -69,6 +69,7 @@
 #include <clocale>
 #include <cstddef>
 #include <string>
+#include <limits>
```
dictoon: Sort alphabetically.

src/appleseed-max-impl/appleseedrenderer/datachunks.h
```diff
@@ -104,6 +106,8 @@ const USHORT ChunkSettingsPostprocessingDenoiseScales               = 0x1507;
 
 const USHORT ChunkSettingsLighting                                  = 0x1600;
 const USHORT ChunkSettingsLightingAlgorithm                         = 0x1601;
+const USHORT ChunkSettingsLightSamplingAlgorithm                    = 0x1602;
+const USHORT ChunkSettingsEnableImportanceSampling                  = 0x1610;
```
dictoon: `ChunkSettingsEnableLightImportanceSampling`

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: Ok, I allow now the full int range, tested the wraparound and it works.

src/appleseed-max-impl/appleseedrenderer/datachunks.h
Mango-3: changed

PR #260, opened 2018-12-26
Add motion blur support

usakhelo:
Hello!

This is quick attempt to add motion blur support. Both object and camera motion blur are supported.
 * To enable object motion blur `appleseed object properties` modifier needs to be added and `optimize instances` should be turned on. Plus Object motion blur should be enabled in standard object properties.
 * Camera motion blur turned on by using Physical camera and by turning on Enable Motion Blur checkbox.
In both cases physical camera's shutter speed is used to determine time offset.

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -1719,14 +1719,20 @@ asf::auto_release_ptr<asr::Camera> build_camera(
 
             if (phys_camera->GetMotionBlurEnabled(time, FOREVER))
             {
-                const float shutter_duration = phys_camera->GetShutterDurationInFrames(time, FOREVER);
-                const int time_offset = asf::round<int, float>(shutter_duration * GetTicksPerFrame());
                 const asf::Transformd transform =
                     asf::Transformd::from_local_to_parent(
                         asf::Matrix4d::make_scaling(asf::Vector3d(settings.m_scale_multiplier)) *
-                        to_matrix4d(view_node->GetObjTMAfterWSM(time + time_offset)));
+                        to_matrix4d(view_node->GetObjTMAfterWSM(time + GetTicksPerFrame())));
 
                 camera->transform_sequence().set_transform(1.0, transform);
+
+                // Set motion blur parameters.
+                const float opening_time = phys_camera->GetShutterOffsetInFrames(time, FOREVER);
+                const float closing_time = opening_time + phys_camera->GetShutterDurationInFrames(time, FOREVER);
+                params.insert("shutter_open_begin_time", opening_time);
+                params.insert("shutter_open_end_time", opening_time);
+                params.insert("shutter_close_begin_time", closing_time);
+                params.insert("shutter_close_end_time", closing_time);
             }
         }
         else
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -819,7 +819,7 @@ namespace
 
                 if (is_animated)
                 {
-                    asf::Transformd animated_transform =
+                   asf::Transformd animated_transform =
                         asf::Transformd::from_local_to_parent(
                             to_matrix4d(node->GetObjTMAfterWSM(time + GetTicksPerFrame())));
 
@@ -1718,12 +1718,13 @@ asf::auto_release_ptr<asr::Camera> build_camera(
                 camera->transform_sequence().set_transform(1.0, transform);
 
                 // Set motion blur parameters.
+                asr::ParamArray& camera_params = camera->get_parameters();
                 const float opening_time = phys_camera->GetShutterOffsetInFrames(time, FOREVER);
                 const float closing_time = opening_time + phys_camera->GetShutterDurationInFrames(time, FOREVER);
-                params.insert("shutter_open_begin_time", opening_time);
-                params.insert("shutter_open_end_time", opening_time);
-                params.insert("shutter_close_begin_time", closing_time);
-                params.insert("shutter_close_end_time", closing_time);
+                camera_params.insert("shutter_open_begin_time", opening_time);
+                camera_params.insert("shutter_open_end_time", opening_time);
+                camera_params.insert("shutter_close_begin_time", closing_time);
+                camera_params.insert("shutter_close_end_time", closing_time);
             }
         }
         else
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -741,6 +741,7 @@ namespace
     void add_object(
         asr::Assembly&          assembly,
         INode*                  node,
+        INode*                  view_node,
         const RenderType        type,
         const RendererSettings& settings,
         const TimeValue         time,
@@ -806,6 +807,43 @@ namespace
             object_assembly_instance->transform_sequence()
                 .set_transform(0.0, transform);
 
+            if (node->GetMotBlurOnOff(time) && node->MotBlur() == 1)
+            {
+                bool is_animated = false;
+                Control* tm_controller;
+                Control* controller;
+                tm_controller = node->GetTMController();
+                auto class_id = tm_controller->ClassID();
+
+                controller = tm_controller->GetPositionController();
+                is_animated = controller->IsAnimated() > 0;
+
+                controller = tm_controller->GetRotationController();
+                is_animated = is_animated || controller->IsAnimated() > 0;
+
+                controller = tm_controller->GetScaleController();
+                is_animated = is_animated || controller->IsAnimated() > 0;
+                
+                if (is_animated)
+                {
+                    float shutter_duration = 0.5f;
+                    if (view_node != nullptr)
+                    {
+                        MaxSDK::IPhysicalCamera* phys_camera = dynamic_cast<MaxSDK::IPhysicalCamera*>(view_node->EvalWorldState(time).obj);
+                        if (phys_camera != nullptr)
+                            shutter_duration = phys_camera->GetShutterDurationInFrames(time, FOREVER);
+                    }
+
+                    int time_offset = asf::round<int, float>(shutter_duration * GetTicksPerFrame());
+                    asf::Transformd transform =
+                        asf::Transformd::from_local_to_parent(
+                            to_matrix4d(node->GetObjTMAfterWSM(time + time_offset)));
+
+                    object_assembly_instance->transform_sequence()
+                        .set_transform(1.0, transform);
+                }
+            }
+
             assembly.assembly_instances().insert(object_assembly_instance);
         }
         else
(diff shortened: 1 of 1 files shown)


Mango-3:
Nice work Sergo!
I think (shutter_duration * ticks_per_frame) should work. In Max all time is in ticks, where 1 second = 4800 ticks. Depending on the frame rate (24, 25 or 30 per second) a frame lasts for
200, 192 or 160 ticks, respectively. You query the shutter duration in number of frames and then multiply with the ticks/frame to get the duration in total number of ticks.
[https://help.autodesk.com/view/3DSMAX/2017/ENU/?guid=__files_GUID_5B1D652F_6BD8_4EAF_8380_4C489363A8C0_htm](url)

usakhelo:
Thanks @Mango-3 ,
I think that's what I did, right? `shutter_duration * GetTicksPerFrame()`
I was so happy they have `GetShutterDurationInFrames` function. In UI max have so many options - frames, time, degrees. But that function just converts everything to frames.

Mango-3:
> I think that's what I did, right? shutter_duration * GetTicksPerFrame()

Yes, that should work, independent of how the shutter time is set (frames, seconds, 1/seconds,..) in the UI, as you sad.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -814,6 +815,43 @@ namespace
             object_assembly_instance->transform_sequence()
                 .set_transform(0.0, transform);
 
+            if (node->GetMotBlurOnOff(time) && node->MotBlur() == 1)
+            {
+                bool is_animated = false;
+                Control* tm_controller;
+                Control* controller;
+                tm_controller = node->GetTMController();
```
dictoon: Initialize at the same time the variable is declared.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -814,6 +815,43 @@ namespace
...
+                auto class_id = tm_controller->ClassID();
+
+                controller = tm_controller->GetPositionController();
```
dictoon: Reusing ("recycling") variables is not ideal.

Much more compact and without variable recycling:

    const bool is_animated =
        tm_controller->GetPositionController()->IsAnimated() > 0 ||
        tm_controller->GetRotationController()->IsAnimated() > 0 ||
        tm_controller->GetScaleController()->IsAnimated() > 0;

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -814,6 +815,43 @@ namespace
...
+                {
+                    float shutter_duration = 0.5f;
+                    if (view_node != nullptr)
+                    {
+                        MaxSDK::IPhysicalCamera* phys_camera = dynamic_cast<MaxSDK::IPhysicalCamera*>(view_node->EvalWorldState(time).obj);
+                        if (phys_camera != nullptr)
+                            shutter_duration = phys_camera->GetShutterDurationInFrames(time, FOREVER);
+                    }
+
+                    int time_offset = asf::round<int, float>(shutter_duration * GetTicksPerFrame());
```
dictoon: Make const.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -814,6 +815,43 @@ namespace
...
+                controller = tm_controller->GetRotationController();
+                is_animated = is_animated || controller->IsAnimated() > 0;
+
+                controller = tm_controller->GetScaleController();
+                is_animated = is_animated || controller->IsAnimated() > 0;
+                
+                if (is_animated)
+                {
+                    float shutter_duration = 0.5f;
+                    if (view_node != nullptr)
```
dictoon: If `view_node` is `nullptr`, `shutter_duration` remains set to 0.5f and so we'll set a transform at time offset `0.5f * GetTicksPerFrame()`. Is that what we want?

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
dictoon: You can remove the explicit `float` type in `asf::round()`.

usakhelo:
Fixed comments and rebased.
Thanks for review, @dictoon!

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
dictoon: I think it's actually not quite right, even when there is a Physical Camera in the scene.

It seems to me that we should always set the camera's transforms at t=0 and t=1, and then we should set the camera's shutter times such that:

- Opening times (begin and end) set to `GetShutterOffsetInFrames()` (or 0.0f when there is no Physical Camera)
- Closing times (begin and end) set to `GetShutterOffsetInFrames() + GetShutterDurationInFrames()` (or 0.5f when there is no Physical Camera).

usakhelo: Hi @dictoon ,
Are 0 and 1 frame numbers? Can I set transform for 1 to `GetObjTMAfterWSM(time + GetTicksPerFrame())`?
You mentioned camera transforms, but line 837 is about assembly transforms. What do we do here? Can I leave this place as is?
I added changes for camera transforms, only for physical camera case. I think only physical camera has Enable Motion Blur checkbox. Other cameras have it as a viewport multi-pass effect. I don't know if we should consider it or should not.

dictoon: Nope, they're arbitrary times. The idea is as follow: you set transforms (for transformation motion blur) or poses (for deformation motion blur) at arbitrary times, be it seconds, frames, etc. Then you tell the camera when to open the shutter, at what speed (and following what curve) and when to close it. The exact unit for the times is not specified by appleseed, it simply needs to be consistent.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1681,6 +1716,24 @@ asf::auto_release_ptr<asr::Camera> build_camera(
...
+            {
+                const asf::Transformd transform =
+                    asf::Transformd::from_local_to_parent(
+                        asf::Matrix4d::make_scaling(asf::Vector3d(settings.m_scale_multiplier)) *
+                        to_matrix4d(view_node->GetObjTMAfterWSM(time + GetTicksPerFrame())));
+
+                camera->transform_sequence().set_transform(1.0, transform);
+
+                // Set motion blur parameters.
+                const float opening_time = phys_camera->GetShutterOffsetInFrames(time, FOREVER);
```
dictoon: This part looks good.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1681,6 +1716,24 @@ asf::auto_release_ptr<asr::Camera> build_camera(
...
+                const float closing_time = opening_time + phys_camera->GetShutterDurationInFrames(time, FOREVER);
+                params.insert("shutter_open_begin_time", opening_time);
```
dictoon: Let's recap:

1. You're setting the first transform at time 0: **since you adopted "frames" as the unit of time**, that means you're setting the first transform at frame 0.

2. You're setting the second transform at time 1, i.e. at frame 1.

3. You're setting the shutter open and close times at (fractional) frame values, which I hope are between 0 and 1.

So it looks correct.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -806,6 +807,36 @@ namespace
...
+                const bool is_animated =
+                    tm_controller->GetPositionController()->IsAnimated() > 0 ||
+                    tm_controller->GetRotationController()->IsAnimated() > 0 ||
+                    tm_controller->GetScaleController()->IsAnimated() > 0;
+
+                if (is_animated)
+                {
+                    float shutter_duration = 0.5f;
+                    if (view_node != nullptr)
+                    {
```
dictoon: The correct thing to do might be to remove everything related to the camera here, and simply set the object's transform at time = 0 (frame 0) and time = 1 (frame 1).

That means you would need to call `GetObjTMAfterWSM()` at `time` and `time + GetTicksPerFrame()`.

dictoon:
This stuff is tricky...

I suggest to compare your results with V-Ray and/or Arnold to make sure we're handling transformation motion blur correctly in all cases.

usakhelo:
Turned out I only have mental ray on my max 2016 and it looks little different than appleseed. I'll test on something more modern if I can get trial/demo access to it.

Mango-3:
You can download Corona for Max (2013-2019) and test it for 45 days free, even for commercial purposes in unlimited demo mode (no resolution limits or watermarks).
[https://corona-renderer.com/download](url)

usakhelo:
Hi @dictoon , I compared last version of object motion blur with Arnold and it looks fine except when open/close duration on the camera is longer than 1 frame.
Transforms on the object are for 0 and 1 frame, right? So if I set shutter duration more than 1 motion blur is broken.
My is guess we still need to check camera shutter duration when calculating transforms for the object and take `max(1.0, shutter_duration)` for the frame 1.

dictoon:
I'm not sure what Arnold does in this case: does it extrapolate the motion outside the [0, 1] range? In appleseed, we assume that the object stops moving before the first transform or after the last, i.e. we clamp motion.

Using `max(1.0, shutter_duration)` for frame 1 would push the time when the object reaches frame 1 in the future, i.e. it would make motion slower, that doesn't seem right (but maybe I'm wrong).

Can you show a comparison between Arnold and appleseed for, say, t=2, t=3 (and maybe t=-1, t=-2) when transforms have been set only at t=0 and t=1?

usakhelo:
Maybe we should just clamp duration to 1.0 then?

Below is the comparison. Lower sphere doesn't have mblur checked in the properties, but arnold ignores that checkbox as long as mblur is checked on camera:

Appleseed duration 2.0:
image

Arnold duration 2.0:
image

Appleseed and arnold duration 1.0:
image
image

usakhelo:
Oh, wait, I guess I misunderstood. Do you mean object have animation keys on frame 0 and 1 and I render on frame 2? There's no motion blur on that frame. I guess I misunderstood again:
Arnold, frame 2, animation of the upper sphere 0-1, duration on top image 1, on other one 2:
image
image

usakhelo:
Hi @dictoon ,
I rebased the branch and made small changes. Right now when `shutter_duration` parameter is 0-1 appleseed rendering looks the same as Arnold. 
When the parameter is >1 Arnold moves the object further in the future, but only until animation keys are set. So if transforms are only set for 0 and 1 frame, and `shutter_duration` = 2, then the blur will be the same as if `shutter_duration` = 1.

dictoon:
I propose to merge this as it looks reasonable and actually matches Arnold for shutter durations in [0, 1].

Once merged we can continue testing and iterate.

Thanks for the quality work!

PR #265, opened 2019-01-30
Expose World Space Velocity diagnostic shader

Mango-3:


usakhelo:
Hi @Mango-3 ,
PR works fine but Regions override doesn't work for some reason. It's says it's not available.

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -166,15 +166,14 @@ const char* get_shader_override_type(const int shader_override_type)
...
-      case 15: return "regions";
-      case 16: return "screen_space_wireframe";
-      case 17: return "shading_normal";
-      case 18: return "sides";
-      case 19: return "tangent";
-      case 20: return "uv";
-      case 21: return "world_space_position";
-      case 22: return "world_space_velocity";
-      case 23: return "world_space_wireframe";
+      case 15: return "screen_space_wireframe";
```
dictoon: This breaks compatibility with existing scenes. Admittedly, it's not a very serious breakage, but the solution is very simple: change

```cpp
case 15: return "regions";

```

to

```cpp
case 15: return "coverage";    // regions are no longer supported
```

What do you think?

Mango-3: I had no problems encountered with the existing scenes I used to test. Maybe if someone had saved the scene with the Regions surface shader enabled it would be remapped now to Screen Space Wireframe (case 15).
But then we would need to keep the UI menu entry, otherwise we would have more cases than surface shader override entries?

est77:
I prefer not to expose world space velocity to plugins and wait for the screen space velocity AOV instead.
It will be confusing to have 2 velocity AOVs, one of them having very little use (if any) in compositing.

dictoon:
@est77 Note that this isn't an AOV (which doesn't exist anyway since I removed it from my PR before we merged it) but a diagnostic mode.

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
dictoon: All diagnostic modes >= 15 would be affected actually, not just the Regions mode.

I think you could do _as if_ you had simply removed the Regions mode, but still remap value 15 ("Regions") to value 6 ("Coverage") or maybe better, 0 ("No Override") when loading the project from disk.

That said I doubt we'll ever hit any problem here so let's ignore this altogether :)

Mango-3:
Ok, I leave it then as is. I don't see a big problem to keep it as a diagnostic shader, in appleseed.studio it is also exposed.

PR #269, opened 2019-02-19
Relabel noise seed controls

Mango-3:
Renaming of the noise seed controls in the UI and enabling variation per frame as default.
Does not change the logic or saved parameters and thus keeps compatibility with saved scenes. 
UI after the changes:
capture


src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -53,7 +53,7 @@ namespace
             m_tile_size = 64;
             m_sampler_type = 0;
             m_noise_seed = 0;
-            m_enable_noise_seed = false;
+            m_enable_noise_seed = true;
```
dictoon: That variable is actually misnamed: it should be `m_enable_noise_seed_variation`, or even better `m_vary_noise_seed`.

Issue #16, opened 2017-02-12
Skip lights with null intensity


dictoon:

dictoon:
Still relevant: we shouldn't even translate lights with zero intensity.

dictoon:
Done.

PR #272, opened 2019-03-17
Skip export of zero intensity lights

Mango-3:
This PR addresses issue #16 
Scene export of light entities with null intensity is avoided.

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -991,7 +991,7 @@ namespace
         const int decay_exponent = light_object->GetDecayType();
 
         // Skip exporting lights with zero intensity
-        if ((color[0] + color[1] + color[2]) > 0.0f && intensity > 0.0f)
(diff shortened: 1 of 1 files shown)


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -990,73 +990,77 @@ namespace
...
-            insert_color(assembly, light_name + "_color", color);
-
-        // Get light from envmap.
-        INode* sun_node(nullptr);
-        BOOL sun_node_on(FALSE);
-        float sun_size_mult;
-        if (rend_params.envMap != nullptr &&
-            rend_params.envMap->IsSubClassOf(AppleseedEnvMap::get_class_id()))
+        // Skip exporting lights with zero intensity
+        if ((color[0] + color[1] + color[2]) > 0.0f && intensity > 0.0f)
```
dictoon: You can use `asf::is_zero()` on colors, spectra, vectors, etc.

Mango-3: Done. Thanks for the tip.

PR #277, opened 2019-03-31
Fix Motion blur on objects attached to animated object

usakhelo:
This PR fixes the bug where children of the animated object don't get any motion blur.
In cases when node is not animated it checks all its parents. If any parent object is animated then motion blur is applied to the object.

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -582,6 +582,15 @@ namespace
         return mtl;
     }
 
+    bool is_node_animated(INode* node)
+    {
+        Control* tm_controller = node->GetTMController();
+        return 
+            tm_controller->GetPositionController()->IsAnimated() > 0 ||
+            tm_controller->GetRotationController()->IsAnimated() > 0 ||
+            tm_controller->GetScaleController()->IsAnimated() > 0;
+    }
+
     enum class RenderType
     {
         Default,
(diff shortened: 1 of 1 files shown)


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -810,16 +819,17 @@ namespace
...
             {
-                Control* tm_controller = node->GetTMController();
-
-                const bool is_animated =
-                    tm_controller->GetPositionController()->IsAnimated() > 0 ||
-                    tm_controller->GetRotationController()->IsAnimated() > 0 ||
-                    tm_controller->GetScaleController()->IsAnimated() > 0;
+                bool is_animated = is_node_animated(node);
+                INode* parent_node = node->GetParentNode();
+                while (!is_animated && !parent_node->IsRootNode())
```
dictoon: I think there's a bug here, you'll never check whether the root node is animated or not.

Here's a solution without any duplication that properly checks the root node:

    bool is_animated = false;

    while (!is_animated)
    {
        is_animated = is_node_animated(node);

        if (node->IsRootNode())
            break;

        node = node->GetParentNode();
    }

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -582,6 +582,15 @@ namespace
         return mtl;
     }
 
+    bool is_node_animated(INode* node)
```
dictoon: Can the argument be made const?

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
usakhelo: Hi @dictoon,
I didn't want to overwrite `node` because it's animation is taken later.
Also I don't think root node can be animated..

dictoon: Alright.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
dictoon: Ping :)

usakhelo: Sorry @dictoon,
No, it can't be const. I tried but it doesn't allow because of the calls it makes inside the function.. Now I'm thinking if I should do the function const?

dictoon: Got it.

It's a free function, not a method, so it cannot be const.

usakhelo: Yes, it doesn't work, `GetTMController()` doesn't allow parameter to be const.

Issue #55, opened 2017-09-01
Allow rotating environment maps


dictoon:

usakhelo:
You can rotate env. maps with offset parameters. Is this about that kind of rotation?

juancarlosgzrz:
Using bitmap options? they are kind of weird. What about supporting dome light?  It makes the job easier.

usakhelo:
Probably it is possible to support standard sphere gizmo. You'll pick the gizmo the same way you choose directional light right now and it binds gizmo's transforms to the map coordinates.

dictoon:
It's already possible to rotate an environment map horizontally and vertically using the U and V offsets:

![image](https://user-images.githubusercontent.com/321290/50113762-2a76bd80-0243-11e9-8e3b-7a6567d06b1a.png)

It would make sense to also allow rotating an environment map using the U/V/W angles:

![image](https://user-images.githubusercontent.com/321290/50113738-1b900b00-0243-11e9-9ba4-00603eefbced.png)

Mango-3:
Fixed with commit a980750

dictoon:
Confirmed done by @juancarlosgzrz.

PR #278, opened 2019-04-01
Fix environment background rotation issues

### LINKED ISSUE #258: V offset is reversed
At least when rotating environment maps, the effect it has on the render is the opposite of the effect it has on the material preview.

Mango-3:
-Aligns the environment map(HDR) orientation with the Max viewport preview
-Allows environment map rotation via the Max Bitmap U,V Offset and U,V Angle controls
-Allows horizontal/vertical flipping via setting the U,V Tiling value ±1.  
-Fixes #258 and #55 

Screenshot shows test render and background preview coinciding (U & V Offset applied).
env_fix

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -1368,25 +1368,25 @@ namespace
                 }
             }
 
-            asf::Vector3d UVFlip_vec(-1,1,1);
+            asf::Vector3d uv_flip_vec(-1.0, 1.0, 1.0);
             UVGen* uvgen = rend_params.envMap->GetTheUVGen();
             if (uvgen && uvgen->IsStdUVGen())
             {
                 StdUVGen* std_uvgen = static_cast<StdUVGen*>(uvgen);
 
-                float h_shift = std_uvgen->GetUOffs(time) * 360.0f + 90.0f + asf::rad_to_deg(std_uvgen->GetUAng(time));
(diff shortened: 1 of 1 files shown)


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1368,19 +1368,33 @@ namespace
                 }
             }
 
+            asf::Vector3d UVFlip_vec(-1,1,1);
```
dictoon: We use snake_case for variable names: `uv_flip_vec`.

Add spaces after commas.

Spell out floating constants: 1.0.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1368,19 +1368,33 @@ namespace
...
             UVGen* uvgen = rend_params.envMap->GetTheUVGen();
             if (uvgen && uvgen->IsStdUVGen())
             {
                 StdUVGen* std_uvgen = static_cast<StdUVGen*>(uvgen);
-                env_edf_params.insert("horizontal_shift", std_uvgen->GetUOffs(time) * 180.0f + 90.0f);
-                env_edf_params.insert("vertical_shift", std_uvgen->GetVOffs(time) * 180.0f);
+
+                float h_shift = std_uvgen->GetUOffs(time) * 360.0f + 90.0f + asf::rad_to_deg(std_uvgen->GetUAng(time));
```
dictoon: Make const.

PR #279, opened 2019-04-09
Collect materials inside root assembly

usakhelo:
Hi,

This change fixes the issue with missing materials when material is assigned to objects in root assembly and to objects in their own assembly (created with appleseed modifier with Optimized for Instancing checked).
Before materials were collected each in the assembly of the object, but in cases of shared material it sometimes gets listed in the nested assembly which makes it invisible to the objects in parent assembly.
This change puts all the materials to the root assembly.

Thanks for review!

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -392,6 +392,7 @@ namespace
 
     MaterialInfo get_or_create_material(
         asr::Assembly&          assembly,
+        asr::Assembly*          root_assembly,
         const std::string&      instance_name,
         Mtl*                    mtl,
         MaterialMap&            material_map,
@@ -399,6 +400,7 @@ namespace
         const TimeValue         time)
     {
         MaterialInfo material_info;
+        asr::Assembly* parent_assembly = root_assembly != nullptr ? root_assembly : &assembly;
 
         auto appleseed_mtl =
             static_cast<IAppleseedMtl*>(mtl->GetInterface(IAppleseedMtl::interface_id()));
@@ -410,10 +412,10 @@ namespace
             {
                 // The appleseed material does not exist yet, let the material plugin create it.
                 material_info.m_name =
-                    make_unique_name(assembly.materials(), wide_to_utf8(mtl->GetName()) + "_mat");
-                assembly.materials().insert(
+                    make_unique_name(parent_assembly->materials(), wide_to_utf8(mtl->GetName()) + "_mat");
+                parent_assembly->materials().insert(
                     appleseed_mtl->create_material(
-                        assembly,
+                        *parent_assembly,
                         material_info.m_name.c_str(),
                         use_max_procedural_maps,
                         time));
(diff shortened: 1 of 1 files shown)


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -410,10 +412,10 @@ namespace
             {
                 // The appleseed material does not exist yet, let the material plugin create it.
                 material_info.m_name =
-                    make_unique_name(assembly.materials(), wide_to_utf8(mtl->GetName()) + "_mat");
-                assembly.materials().insert(
+                    make_unique_name(parent_assembly->materials(), wide_to_utf8(mtl->GetName()) + "_mat");
+                parent_assembly->materials().insert(
```
dictoon: I guess it's intentional that when a material is created, it's created in the "parent assembly" and not in `assembly` since it allows more sharing, is that correct?

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -785,6 +790,7 @@ namespace
                 {
                     create_object_instance(
                         object_assembly.ref(),
+                        &assembly,
```
dictoon: An alternative idea would have been the following:

In appleseed, entities are "parented": you can retrieve the parent of an entity using `get_parent()`. An entity is assigned a parent when it gets inserted into a collection.

What we could have done is manually set the parent of `object_assembly` to be `assembly`. Then `get_or_create_material()` would have been able to walk the chain of parents up until it found the material it was looking for. When `object_assembly` would then be inserted into `assembly.assemblies()`, it's parent would be overwritten (to the same value).

However it would then be unclear in which assembly should `get_or_create_material()` create the material. So all in all not the best solution probably, but I just wanted to share the idea.

usakhelo: Hi @dictoon,
I tried suggested approach first, but it didn't work because objects assemblies get added to patent assemblies later. At the point when assembly is populated with materials and objects it doesn't have a parent.

dictoon: Yes, that's why I was suggesting to manually parent the entity first using `set_parent()` but overall I prefer your approach.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
dictoon: If you confirm this then we can merge the PR.

usakhelo: Yes, parent_assembly should be root assembly in cases of nested assembly.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
usakhelo: I tried to parent assembly first and then populate it but it started to crash. I thought it's not designed to be filled out once it's inserted into parent assembly.

dictoon:
Thanks!

PR #283, opened 2019-04-16
Fix VS solutions

usakhelo:
Hi,
I had to make some changes to solution files and to some appleseed calls in order to compile against recent appleseed master. I changed windows sdk according to 3ds max SDK requirements and changes paths to deps and stage folders.
I remember I saw some discussion on deps folder naming, but couldn't recall what was the conclusion.

Thanks,
Sergo.


dictoon:
Yeah, we're a bit in the middle of a unification here (unifying `vs140`, `vc14` and `v140` as `vc140`) and it's my fault for not communicating clearly about it and for having missed it in my review.

To summarize:

- In _downloadable_ appleseed-deps packages, the folders are called `vc140` and `vc141`.
- In appleseed-deps's `BuildAll.bat`, the folders are called `vc14` and `vc141`.
- In the sandbox, the folders are called `vc14` and `vc141`.

I'll update `BuildAll.bat` in a minute to use `vc140` instead of `vc14`.

Should we bite the bullet and complete the transition, that is, rename the folders in the sandbox? In which case this PR should no longer be useful.

dictoon:
Here's appleseed-deps' `BuildAll.bat` update:
https://github.com/appleseedhq/appleseed-deps/pull/5

@usakhelo I realized that this PR also contains code updates. Would you mind opening a separate PR with just those?

dictoon:
And here's an update for the folders in the sandbox, it was a small change:
https://github.com/appleseedhq/appleseed/pull/2557

usakhelo:
Thanks @dictoon , I'll do it.

usakhelo:
@dictoon PR is ready: https://github.com/appleseedhq/appleseed-max/pull/284
@Mango-3 did you compile with the recent appleseed master? I couldn't compile it without those changes.

Mango-3:
-> did you compile with the recent appleseed master? I couldn't compile it without those changes.
When I made my last PR with the Visual Studio solution files it was compiling fine without those. But since then master had changed, requiring your modifications.

usakhelo:
Got it, thanks.
Closing this PR.

PR #292, opened 2019-05-14
Add build options include

usakhelo:
Include `buildoptions.h` header to avoid link errors.


Mango-3:
Looking good. The only thing missing is to remove the preprocessor definitions also in the project files for Max 2017, 2018 & 2020 as you did for 2019. 
I made a test build for Max 2019 and had no problems.

usakhelo:
Good point @Mango-3 ,
I added the commit and checked if 2017 compiles and it does. I can't check 2018 and 2020.

dictoon:
I will check.

dictoon:
Builds on all 3ds Max version we support (2017-2020).

dictoon:
Thanks!

PR #295, opened 2019-05-20
Undefine base_type macro to avoid conflict with boost type

usakhelo:
This is quick PR that removes base_type macro from plugin files. It helps to prevent boost error message during compilation that started to appear in recent build.

### DIFF
--- a/src/appleseed-max-impl/_beginmaxheaders.h
+++ b/src/appleseed-max-impl/_beginmaxheaders.h
@@ -5,7 +5,7 @@
 //
 // This software is released under the MIT license.
 //
-// Copyright (c) 2016-2018 Francois Beaune, The appleseedhq Organization
+// Copyright (c) 2015-2019 Sergo Pogosyan, The appleseedhq Organization
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -27,26 +27,3 @@
 //
 
 #pragma once
-
-// Build options header.
-#include "renderer/api/buildoptions.h"
-
-// appleseed.foundation headers.
-#include "foundation/utility/log.h"
-
-// Standard headers.
-#include <cstddef>
-
-class LogTarget
-  : public foundation::ILogTarget
-{
-  public:
-    void release() override;
-
-    void write(
-        const foundation::LogMessage::Category  category,
-        const char*                             file,
-        const size_t                            line,
-        const char*                             header,
-        const char*                             message) override;
-};
--- a/src/appleseed-max-impl/_beginmaxheaders.h
+++ b/src/appleseed-max-impl/_beginmaxheaders.h
@@ -5,7 +5,7 @@
 //
 // This software is released under the MIT license.
 //
-// Copyright (c) 2015-2019 Sergo Pogosyan, The appleseedhq Organization
+// Copyright (c) 2019 Sergo Pogosyan, The appleseedhq Organization
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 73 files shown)


src/appleseed-max-impl/logtarget.h
```diff
@@ -36,6 +36,7 @@
 
 // Standard headers.
 #include <cstddef>
+#undef base_type
```
dictoon: It seems redundant here.

usakhelo:
Begin/end header files added. @dictoon could you check if I understood correctly?

src/appleseed-max-impl/_beginmaxheaders.h
```diff
@@ -0,0 +1,29 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2015-2019 Sergo Pogosyan, The appleseedhq Organization
```
dictoon: Just 2019.

src/appleseed-max-impl/_endmaxheaders.h
```diff
@@ -0,0 +1,31 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2015-2019 Sergo Pogosyan, The appleseedhq Organization
```
dictoon: Same here.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -776,7 +782,7 @@ namespace
             asf::Transformd::from_local_to_parent(
                 to_matrix4d(node->GetObjTMAfterWSM(time)));
 
-        if (should_optimize_for_instancing(object, time))
+        if (motion_blur_enabled(node, time) || should_optimize_for_instancing(object, time))
```
dictoon: Very nice, but unrelated to this PR?

src/appleseed-max-impl/utilities.h
```diff
@@ -54,6 +54,7 @@
 #include <maxtypes.h>
 #include <point3.h>
 #include <point4.h>
+#undef base_type
```
dictoon: Fix here too.

In fact, it seems to me that there are many other places where Max files are included but are missing the header guards.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
usakhelo: Sorry, that's part of other pr, right

src/appleseed-max-impl/utilities.h
usakhelo: I'm adding guards in .h files. Should I do the same for .cpp? Looks like it, they also include max headers _after_ their own .h files..

usakhelo: But because of `#pragma once` it won't do anything in .cpp file, right? Should I remove `#pragma` statement?

src/appleseed-max-impl/appleseedrenderer/dialoglogtarget.h
```diff
@@ -35,10 +35,12 @@
 #include "foundation/utility/log.h"
 
 // Standard headers.
+#include "_beginmaxheaders.h"
```
dictoon: They shouldn't be added here.

src/appleseed-max-impl/builtinmapsupport.cpp
```diff
@@ -33,6 +33,7 @@
 #include "appleseedoslplugin/osltexture.h"
 #include "oslutils.h"
 #include "utilities.h"
+#include "_endmaxheaders.h"
```
dictoon: Something's wrong here, remove.

PR #297, opened 2019-05-22
Add support for geometric object plugins, many cleanups

dictoon:
Depends on https://github.com/appleseedhq/appleseed/pull/2626.


src/_beginmaxheaders.h
```diff
@@ -25,3 +25,9 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 //
+
+// Build options header.
+#include "foundation/core/buildoptions.h"
```
usakhelo: I like these two headers included here but this header still needs to be included separately in many places.

usakhelo:
I like how cleaner it become, thanks @dictoon . It compiles and runs fine, checked on few of my test scenes.

src/_beginmaxheaders.h
dictoon: As far as I can tell it's included everywhere where it must be, can you point a place where that isn't the case?

usakhelo: No @dictoon, it's all fine. It's just that _begin header little later includes it again. But that's not big deal. It was just a thought, sorry.

dictoon: No problem, thanks for the review!

PR #300, opened 2019-06-16
Support meshes without smoothing groups

usakhelo:
Hi,
I had an issue where meshes without smoothing groups were not rendered properly. This PR should fix the issue. Attached is the test scene with different meshes.
[test_normal_export.zip](https://github.com/appleseedhq/appleseed-max/files/3293676/test_normal_export.zip)

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
@@ -82,6 +82,7 @@
 #include <genlight.h>
 #include <iInstanceMgr.h>
 #include <INodeTab.h>
+#include <MeshNormalSpec.h>
 #include <modstack.h>
 #include <object.h>
 #include <pbbitmap.h>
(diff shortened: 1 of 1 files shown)


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -231,15 +232,38 @@ namespace
...
+                        normal_indices[j] =
+                            static_cast<asf::uint32>(
+                                object->push_vertex_normal(
+                                    asf::safe_normalize(asr::GVector3(n.x, n.y, n.z))));
+                    }
+                }
+
+                if (!normal_set)
+                {
+                    // No explicit normals for this face, use face normal
```
dictoon: End comment with a dot.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -231,15 +232,38 @@ namespace
...
-                normal_indices[2] = normal_index;
+                // No smooth group for this face, check if face has explicit normals.
+                bool normal_set = false;
+                MeshNormalSpec* nspec = mesh.GetSpecifiedNormals();
+                if (nspec)
+                {
+                    for (int j = 0; j < 3; ++j)
+                    {
+                        const int norm_index = nspec->GetNormalIndex(i, j);
+                        if (norm_index < 0)
```
dictoon: I feel like there's a potential bug here:

Can it be that there's a normal at j == 0, in which case `normal_set` is set to true, but no normal at j == 1, in which case we break out of the loop (but `normal_set` is set to true so we won't switch to using the face normal)?

usakhelo: I think I fixed it @dictoon

dictoon:
Needs to be rebased and squashed.

PR #301, opened 2019-07-07
Remove obsolete pixel filter methods

Mango-3:
- Removes pixel filter which had been removed from appleseed-core when filter importance sampling was introduced. 

The removed legacy filter are all mapped now to Blackman-Harris, the supported pixel filter mappings are unaffected. Removed pixel filter will show up in the UI marked legacy (see screenshot)
Capture


src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1725,15 +1725,15 @@ namespace
...
-          case 0:  return "blackman-harris";
-          case 1:  return "box";
-          case 2:  return "catmull";
-          case 3:  return "bspline";
-          case 4:  return "gaussian";
-          case 5:  return "lanczos";
-          case 6:  return "mitchell";
-          case 7:  return "triangle";
-          default: return "box";
+        case 0:  return "blackman-harris";
```
usakhelo: I think two space indentation is missing here.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
```diff
@@ -195,11 +195,11 @@ namespace
     {
         { IDS_RENDERERPARAMS_FILTER_TYPE_1,             L"Blackman-Harris" },
         { IDS_RENDERERPARAMS_FILTER_TYPE_2,             L"Box" },
-        { IDS_RENDERERPARAMS_FILTER_TYPE_3,             L"Catmull-Rom Spline" },
-        { IDS_RENDERERPARAMS_FILTER_TYPE_4,             L"Cubic B-spline" },
+        { IDS_RENDERERPARAMS_FILTER_TYPE_3,             L" Catmull-Rom Spline - Legacy" },
```
usakhelo: One space indentation is deliberate I suppose?

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
Mango-3: True, the auto indent struck again.

src/appleseed-max-impl/appleseedrenderer/appleseedrenderer.cpp
Mango-3: It is intentional. My idea was to have the legacy filter entries slightly offset in the selection menu (see screenshot above) to make it visually more distinct.

src/appleseed-max-impl/appleseedrenderer/projectbuilder.cpp
```diff
@@ -1727,13 +1727,13 @@ namespace
         {
           case 0:  return "blackman-harris";
           case 1:  return "box";
-          case 2:  return "catmull";
-          case 3:  return "bspline";
+          case 2:  return "blackman-harris"; // legacy catmul mapped to blackman-harris
```
dictoon: Typo: *catmull*

Mango-3:
> So if I understand correctly, on an older scene that uses one of the removed filter, the filter combo will show the name of that removed filter (with a "legacy" tag) but under the hood it will use Blackman-Harris

Yes, this is the case. It keeps the order of the supported pixel filter intact when loading legacy scene files. The obsolete ones are all remapped to Blackman-Harris. 
The legacy tag should user discourage using it for new scenes.

PR #302, opened 2019-08-03
Disable Light Tree in UI and bump version for release

Mango-3:


dictoon:
On second thought, maybe we should also remove `IDS_RENDERERPARAMS_LIGHT_SAMPLER_TYPE_2` and the associated string. That way, we avoid having unused resources. We can always add them back later on.

Mango-3:
Done

dictoon:
Thanks!

PR #306, opened 2019-08-18
Add appleseed-max-common project, switch to property sheets, allow assigning materials to non-mesh objects

dictoon:


Mango-3:
Compiles and works without problems for Max 2020

dictoon:
Thanks guys!

Issue #32, opened 2017-06-16
Add exposure to appleseed light material


dictoon:

juancarlosgzrz:
I think this can be closed?

dictoon:
Yes, thanks for the heads up!

PR #309, opened 2019-08-30
Add exposure parameter to light material

jjeziorski:
Here's a simple PR for https://github.com/appleseedhq/appleseed-max/issues/32

### DIFF
--- a/src/appleseed-max-impl/appleseedlightmtl/appleseedlightmtl.cpp
+++ b/src/appleseed-max-impl/appleseedlightmtl/appleseedlightmtl.cpp
@@ -94,7 +94,8 @@ namespace
         ParamIdLightColorTexmap = 1,
         ParamIdLightPower       = 2,
         ParamIdEmissionFront    = 3,
-        ParamIdEmissionBack     = 4
+        ParamIdEmissionBack     = 4,
+        ParamIdLightExposure    = 5
     };
 
     enum TexmapId
@@ -147,6 +148,12 @@ namespace
             p_default, 1.0f,
             p_range, 0.0f, 1000000.0f,
             p_ui, TYPE_SPINNER, EDITTYPE_FLOAT, IDC_EDIT_LIGHT_POWER, IDC_SPINNER_LIGHT_POWER, SPIN_AUTOSCALE,
+        p_end,
+
+        ParamIdLightExposure, L"light_exposure", TYPE_FLOAT, P_ANIMATABLE, IDS_LIGHT_EXPOSURE,
+            p_default, 0.0f,
+            p_range, 0.0f, 100.0f,
+            p_ui, TYPE_SPINNER, EDITTYPE_FLOAT, IDC_EDIT_LIGHT_EXPOSURE, IDC_SPINNER_LIGHT_EXPOSURE, SPIN_AUTOSCALE,
         p_end,
 
         ParamIdEmissionFront, L"emission_front", TYPE_BOOL, 0, IDS_EMISSION_FRONT,
@@ -335,6 +342,9 @@ void AppleseedLightMtl::Update(TimeValue t, Interval& valid)
         m_pblock->GetValue(ParamIdLightColorTexmap, t, m_light_color_texmap, m_params_validity);
 
         m_pblock->GetValue(ParamIdLightPower, t, m_light_power, m_params_validity);
+        float exposure;
+        m_pblock->GetValue(ParamIdLightExposure, t, exposure, m_params_validity);
+        m_light_power *= std::pow(2.0f, exposure);
 
         int emission_front;
         m_pblock->GetValue(ParamIdEmissionFront, t, emission_front, m_params_validity);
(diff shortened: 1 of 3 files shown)


usakhelo:
Hi @jjeziorski ,
This parameter is the synonym of `m_light_power` right? It's fine I think. 
Have you tried to open scene saved with previous version of the plugin? We are trying to maintain backward compatibility.

jjeziorski:
Hi @usakhelo,
Thank you for review.
I'm not sure if I'd call it synonym of `m_light_power`, but indeed it modifies light power just using exponential scale.

Yes, I've checked opening old scene and it works for me. Anyway as far as I know, if you keep paramblock definition consistent (ids / types) than scenes are always backward compatible.

usakhelo:
@jjeziorski this PR should be rebased. Would be great if @dictoon review this. If not let's merge this too, it looks fine to me.

jjeziorski:
@usakhelo I don't see any conflicts, so rebase should be trivial upon merge (github option).
Having said that, please let me know if there are any conflicts/issues.

src/appleseed-max-impl/appleseedlightmtl/appleseedlightmtl.cpp
```diff
@@ -149,6 +150,12 @@ namespace
             p_ui, TYPE_SPINNER, EDITTYPE_FLOAT, IDC_EDIT_LIGHT_POWER, IDC_SPINNER_LIGHT_POWER, SPIN_AUTOSCALE,
         p_end,
 
+        ParamIdLightExposure, L"light_exposure", TYPE_FLOAT, P_ANIMATABLE, IDS_LIGHT_EXPOSURE,
+            p_default, 0.0f,
+            p_range, 0.0f, 100.0f,
```
dictoon: The range should allow negative values. We usually use [-64, 64] as exposure range.

src/appleseed-max-impl/appleseedlightmtl/appleseedlightmtl.cpp
```diff
@@ -149,6 +150,12 @@ namespace
             p_ui, TYPE_SPINNER, EDITTYPE_FLOAT, IDC_EDIT_LIGHT_POWER, IDC_SPINNER_LIGHT_POWER, SPIN_AUTOSCALE,
         p_end,
 
+        ParamIdLightExposure, L"light_exposure", TYPE_FLOAT, P_ANIMATABLE, IDS_LIGHT_EXPOSURE,
+            p_default, 0.0f,
```
dictoon: Default should be 1.

src/appleseed-max-impl/appleseedlightmtl/appleseedlightmtl.cpp
```diff
@@ -335,6 +342,9 @@ void AppleseedLightMtl::Update(TimeValue t, Interval& valid)
         m_pblock->GetValue(ParamIdLightColorTexmap, t, m_light_color_texmap, m_params_validity);
 
         m_pblock->GetValue(ParamIdLightPower, t, m_light_power, m_params_validity);
+        float exposure;
```
dictoon: Add blank line between lines 344 and 345 to separate blocks of related lines.

src/appleseed-max-impl/appleseedlightmtl/appleseedlightmtl.cpp
```diff
@@ -335,6 +342,9 @@ void AppleseedLightMtl::Update(TimeValue t, Interval& valid)
...
+        m_pblock->GetValue(ParamIdLightExposure, t, exposure, m_params_validity);
+        m_light_power *= std::pow(2.0f, exposure);
```
dictoon: You need to include `<cmath>`.

dictoon: Also, side note: the `DiffuseEDF` model actually has an `exposure` parameter. The upside of using it would be that it can be textured, but since we don't seem to allow this in this material, the way you did it is fine.

dictoon:
Thanks for the contribution @jjeziorski!

usakhelo:
Thanks @dictoon, I feel much better about this PR now :)

usakhelo:
@jjeziorski it should rebase with no issues, you are right.

src/appleseed-max-impl/appleseedlightmtl/appleseedlightmtl.cpp
jjeziorski: I don't think it's a good idea. This would make all scenes previously setup with `appleseedlightmtl` twice as bright. Also value of 1 seems rather arbitrary, value of 0 is neutral and as such has special meaning.
Having said that, if you feel this is still needed for consistency or otherwise, I'll gladly change that.

dictoon: Doh, of course you're right. Facepalm :)

src/appleseed-max-impl/appleseedlightmtl/appleseedlightmtl.cpp
dictoon: Actually, now that I think of it, it would be nicer to use the `exposure` parameter natively supported by the EDF so that it appears as such when the scenes are opened in appleseed.studio. Can be changed in a separate PR.

Issue #14, opened 2016-10-15
Add alpha to SSS material


dictoon:

dictoon:
@juancarlosgzrz In your opinion, is this issue still relevant? Is there currently a way to have both SSS and alpha transparency/cutout on an object, say a tree leaf?

dictoon:
In any case, if we decide to add an alpha parameter, it should be called `Alpha` and vary from 0 to 100 (percents).

Another point to check is whether we want to update the built-in shader (not based on OSL), or the OSL shader (the one you pointed out), or both.

juancarlosgzrz:
> @juancarlosgzrz In your opinion, is this issue still relevant? Is there currently a way to have both SSS and alpha transparency/cutout on an object, say a tree leaf?

Is still relevant, and there's no workaround, in that case I would blend two shaders together, the leaf material and a transparent material and mask them, but that way creates more unnecessary and complex shader trees.  

I would vote yes

dictoon:
Done.

PR #315, opened 2019-09-02
Introduce alpha parameter to SSS material

jjeziorski:
Here's a PR for this issue https://github.com/appleseedhq/appleseed-max/issues/14

It seems to work fine, but I have two questions:
1. Where should it be located in UI? At the moment it's in SSS tab, whereas it applies to both SSS and Specular, which might be misleading. However a separate tab for single map seems excessive. Maybe it should be added to bump tab? Though in that case it would need to be removed from Disney material tab. I'm not sure what is the best approach here.
2. It does not work well with appleseed texture map, whereas it works fine with native Max bitmap node. Is it expected?


dictoon:
About the UI: I think the "SSS Parameters" section is appropriate. It's consistent with how things are done in the Disney material.

About the appleseed texture map problem: nope, this isn't expected. What's the issue exactly?

In any case, this PR is pretty much perfect, thanks for the quality code, will merge now!

PR #332, opened 2019-10-29
Expose shadow terminator correction factor

Mango-3:
Implements a control for changing the shadow terminator correction factor within the appleseed object modifier.

Depends on https://github.com/appleseedhq/appleseed/pull/2720


src/appleseed-max-impl/appleseedobjpropsmod/appleseedobjpropsmod.cpp
```diff
@@ -153,6 +155,10 @@ namespace
             p_default, FALSE,
             p_ui, ParamMapIdVisibility, TYPE_SINGLECHECKBOX, IDC_BUTTON_PHOTON_TARGET,
         p_end,
+        ParamIdShadowTerminatorCorrection, L"shadow_terminator_correction", TYPE_FLOAT, P_TRANSIENT, 0,
+            p_ui, ParamMapIdVisibility, TYPE_SPINNER, EDITTYPE_FLOAT, IDS_TEXT_SHADOW_TERMINATOR_CORRECTION, IDS_SPINNER_SHADOW_TERMINATOR_CORRECTION, SPIN_AUTOSCALE,
+            p_default, 0.0f, p_range, 0.0f, 0.5f,
```
dictoon: Are these soft or hard bounds?

src/appleseed-max-impl/appleseedobjpropsmod/appleseedobjpropsmod.cpp
```diff
@@ -365,6 +371,10 @@ int AppleseedObjPropsMod::get_medium_priority(const TimeValue t) const
    return m_pblock->GetInt(ParamIdMediumPriority, t, FOREVER);
 }
 
+float AppleseedObjPropsMod::get_shadow_terminator_correction(const TimeValue t) const
+{
+    return m_pblock->GetFloat(ParamIdShadowTerminatorCorrection, t, FOREVER);
+}
```
dictoon: Keep two blank lines before block comments.

src/appleseed-max-impl/appleseedobjpropsmod/appleseedobjpropsmod.cpp
Mango-3: Hard bounds. Max does not have the soft bounds like Maya has. I have seen that also in appleseed.studio the upper limit has been set to 0.5f (hard limit).

dictoon:
Thanks!

PR #335, opened 2019-11-25
Fix issues in OSL plugin

usakhelo:
Hi,
These are changes that fix some of the hidden issues OSL plugin:
1. Add Version information to the param block
2. Skip parameters with obsolete flag
3. Add separate variable to keep param ids for maps
4. Start bump parameters IDs with 0 (because they are in separate param block) This may potentially break current scenes but it worked on my tests.
5. Skip parameters with no widget and no connections (Don't remember where this came from but it sounds reasonable. I probably had a case where it failed to add the parameter)
6. Fix getting correct colors from OSL color default values.

Thanks for review!

### DIFF
--- a/src/appleseed-max-impl/appleseedoslplugin/oslshadermetadata.cpp
+++ b/src/appleseed-max-impl/appleseedoslplugin/oslshadermetadata.cpp
@@ -238,15 +238,19 @@ OSLShaderInfo::OSLShaderInfo(
         {
             OSLParamInfo osl_param(q.get_param_info(i));
 
+            if (osl_param.m_widget == "null" && !osl_param.m_connectable)
+                continue;
+
             MaxParam& max_param = osl_param.m_max_param;
 
             max_param.m_osl_param_name = osl_param.m_param_name;
             max_param.m_max_label_str = osl_param.m_label;
-            max_param.m_connectable = osl_param.m_connectable;
+            max_param.m_is_connectable = osl_param.m_connectable;
             max_param.m_param_type = MaxParam::Unsupported;
             max_param.m_page_name = osl_param.m_page;
 
-            max_param.m_has_constant = osl_param.m_valid_default &&
+            max_param.m_is_constant =
+                osl_param.m_valid_default &&
                 osl_param.m_lock_geom &&
                 osl_param.m_widget != "null";
 
--- a/src/appleseed-max-impl/appleseedoslplugin/oslshadermetadata.h
+++ b/src/appleseed-max-impl/appleseedoslplugin/oslshadermetadata.h
@@ -82,10 +82,11 @@ class MaxParam
     }
 
     ParamType       m_param_type;
-    bool            m_connectable;
-    bool            m_has_constant;
+    bool            m_is_connectable;
+    bool            m_is_constant;
     std::string     m_max_label_str;
     int             m_max_param_id;
+    int             m_max_map_param_id;
     int             m_max_ctrl_id;
     std::string     m_osl_param_name;
     std::string     m_page_name;
--- a/src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
+++ b/src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
@@ -61,6 +61,7 @@
 #include "boost/filesystem.hpp"
 
 // Standard headers.
+#include <map>
 #include <memory>
 
 namespace asf = foundation;
@@ -338,9 +339,10 @@ namespace
 
     void add_bump_parameters(
         ParamBlockDesc2*            pb_desc,
-        IdNameVector&               texture_map,
-        int&                        param_id)
+        IdNameVector&               texture_map)
     {
+        int param_id = 0;
+
         pb_desc->AddParam(
             param_id++, L"bump_method", TYPE_INT, 0, IDS_BUMP_METHOD,
             p_ui, TYPE_INT_COMBOBOX, IDC_COMBO_BUMP_METHOD,
@@ -426,8 +428,8 @@ void OSLShaderRegistry::create_class_descriptors()
             L"oslTextureMapParams",                     // internal parameter block's name
             0,                                          // ID of the localized name string
             class_descr,                                // class descriptor
-            P_AUTO_CONSTRUCT,                           // block flags
-
+            P_AUTO_CONSTRUCT | P_VERSION,               // block flags
+            1,                                          // P_VERSION argument
                                                         // --- P_AUTO_CONSTRUCT arguments ---
             0,                                          // parameter block's reference number
             p_end
@@ -439,12 +441,12 @@ void OSLShaderRegistry::create_class_descriptors()
         for (auto& param_info : shader.m_params)
         {
             param_info.m_max_param.m_max_ctrl_id = ctrl_id++;
-            param_info.m_max_param.m_max_param_id = param_id;
 
-            shader.m_string_map.insert(std::make_pair(string_id, utf8_to_wide(param_info.m_max_param.m_max_label_str)));
-
-            if (param_info.m_max_param.m_has_constant)
+            if (param_info.m_max_param.m_is_constant)
             {
+                param_info.m_max_param.m_max_param_id = param_id;
+
+                shader.m_string_map.insert(std::make_pair(string_id, utf8_to_wide(param_info.m_max_param.m_max_label_str)));
                 add_const_parameter(
                     param_block_descr,
                     param_info,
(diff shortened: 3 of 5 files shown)


src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
```diff
@@ -426,8 +428,9 @@ void OSLShaderRegistry::create_class_descriptors()
             L"oslTextureMapParams",                     // internal parameter block's name
             0,                                          // ID of the localized name string
             class_descr,                                // class descriptor
-            P_AUTO_CONSTRUCT,                           // block flags
-
+            P_AUTO_CONSTRUCT | P_VERSION,               // block flags
+                                                        // --- P_VERSION arguments ---
+            1,
```
dictoon: That's supposed to be on the previous line no?

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
```diff
@@ -439,12 +442,12 @@ void OSLShaderRegistry::create_class_descriptors()
         for (auto& param_info : shader.m_params)
         {
             param_info.m_max_param.m_max_ctrl_id = ctrl_id++;
-            param_info.m_max_param.m_max_param_id = param_id;
-
-            shader.m_string_map.insert(std::make_pair(string_id, utf8_to_wide(param_info.m_max_param.m_max_label_str)));
 
             if (param_info.m_max_param.m_has_constant)
```
dictoon: Unrelated to this PR but I wonder: what does "has constant" mean? Should it be `m_is_constant`? (In which case it should also be `m_is_connectable`, etc.)

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
```diff
@@ -546,8 +553,8 @@ void OSLShaderRegistry::add_const_parameter(
         if (osl_param.m_has_default)
         {
             const float r = static_cast<float>(osl_param.m_default_value.at(0));
-            const float g = static_cast<float>(osl_param.m_default_value.at(0));
-            const float b = static_cast<float>(osl_param.m_default_value.at(0));
+            const float g = static_cast<float>(osl_param.m_default_value.at(1));
```
dictoon: That looks like a bug fix? In which case it deserves its own commit.

src/appleseed-max-impl/appleseedoslplugin/oslshadermetadata.h
```diff
@@ -86,6 +86,7 @@ class MaxParam
     bool            m_has_constant;
     std::string     m_max_label_str;
     int             m_max_param_id;
+    int             m_max_map_param_id;
```
dictoon: What does the word "map" mean in this context?

usakhelo: Map is a texture input. Before id of a texture input was calculated based on float or color parameter id, now it's separate variable.

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
usakhelo: It is bug fix yes. I'll do separate commit for this.

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
usakhelo: Constant is color or float parameter, which can or cannot be textured. I can rename it to m_is_constant

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
usakhelo: Probably, since it's only one argument...

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
usakhelo: @dictoon just the commit or separate PR for this?

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
usakhelo: Now I think `m_is_constant` is misleading because parameter can be a constant and be connectable.. Is it ok you think to have both `m_is_constant` and `m_is_connectable` set to true?

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
dictoon: A separate commit is enough!

src/appleseed-max-impl/appleseedoslplugin/oslshadermetadata.h
dictoon: Oh ok, makes sense, thanks.

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
dictoon: I was just wondering about the choice of verb ("has" vs "is"). I don't know the code enough to know which verb is more accurate.

src/appleseed-max-impl/appleseedoslplugin/oslshadermetadata.cpp
```diff
@@ -249,7 +249,7 @@ OSLShaderInfo::OSLShaderInfo(
             max_param.m_param_type = MaxParam::Unsupported;
             max_param.m_page_name = osl_param.m_page;
 
-            max_param.m_has_constant = osl_param.m_valid_default &&
+            max_param.m_is_constant = osl_param.m_valid_default &&
```
dictoon: Formatting suggestion:

    max_param.m_is_constant =
        osl_param.m_valid_default &&
        osl_param.m_lock_geom &&
        osl_param.m_widget != "null";

src/appleseed-max-impl/appleseedoslplugin/oslshadermetadata.h
```diff
@@ -83,7 +83,7 @@ class MaxParam
 
     ParamType       m_param_type;
     bool            m_connectable;
```
dictoon: This should be renamed to `m_is_connectable` (if it makes sense).

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
```diff
@@ -821,7 +820,7 @@ void OSLShaderRegistry::add_input_parameter(
         auto param_str = utf8_to_wide(max_param.m_osl_param_name + "_map");
         texture_map.push_back(std::make_pair(param_id, utf8_to_wide(max_param.m_max_label_str)));
 
-        const bool short_button = max_param.m_has_constant &&
+        const bool short_button = max_param.m_is_constant &&
```
dictoon: Formatting suggestion:

    const bool short_button =
        max_param.m_is_constant &&
        (max_param.m_param_type == MaxParam::VectorParam ||
         max_param.m_param_type == MaxParam::NormalParam ||
         max_param.m_param_type == MaxParam::PointParam ||
         max_param.m_param_type == MaxParam::StringPopup ||
         max_param.m_param_type == MaxParam::IntMapper);

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
```diff
@@ -460,6 +462,9 @@ void OSLShaderRegistry::create_class_descriptors()
 
             if (param_info.m_max_param.m_connectable)
```
Mango-3: Franz mentioned this before, better to rename it to `m_is_connectable`

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
```diff
@@ -486,7 +491,7 @@ void OSLShaderRegistry::create_class_descriptors()
         max_output_param.m_max_ctrl_id = ctrl_id++;
         max_output_param.m_max_param_id = param_id;
         max_output_param.m_connectable = false;
```
Mango-3: Same as above

src/appleseed-max-impl/appleseedoslplugin/oslshaderregistry.cpp
```diff
@@ -62,6 +62,7 @@
 
 // Standard headers.
 #include <memory>
+#include <map>
```
Mango-3: Order alphabetically

PR #337, opened 2019-11-29
Change custom fixed width integer types to C++11 std ones

Mango-3:
- Changes obsolete appleseed-foundation custom fixed width integer types with C++11 compatible ones. 
Related to [https://github.com/appleseedhq/appleseed/pull/2743](url)

### DIFF
--- a/src/appleseed-max-impl/utilities.cpp
+++ b/src/appleseed-max-impl/utilities.cpp
@@ -582,7 +582,7 @@ namespace
         {
         }
 
-        asf::uint64 compute_signature() const override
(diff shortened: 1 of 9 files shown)


src/appleseed-max-impl/utilities.cpp
```diff
@@ -582,7 +582,7 @@ namespace
         {
         }
 
-        asf::uint64 compute_signature() const override
```
dictoon: Include `<cstdint>`?

dictoon: I'll take care of this.

PR #338, opened 2019-11-30
Minor cleanups

dictoon:


Mango-3:
Looks good to me. 
The [Max SDK docs state](http://help.autodesk.com/view/3DSMAX/2020/ENU/?guid=__cpp_ref_group___assert_macros_html) that assert and DbgAssert are equivalent and Plug-ins are encouraged to use the DbgAssert macro.

dictoon:
Thanks for digging into the docs! I'm somewhat shocked that they dared redefine `assert()`...

Issue #314, opened 2019-09-01
appleseed-maya defaults to -1 rendering threads, but appleseed-max defaults to 0


dictoon:
Would it make sense to switch appleseed-max to `-1` as well by default?

Reminder:
- `0` means "use all cores of the machine"
- `-1` means "use all cores of the machine but 1"

dictoon:
Done.

PR #339, opened 2019-11-30
Changes default render threads from all to all - 1

Mango-3:
- Implements request #314

### DIFF
--- a/src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
+++ b/src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
@@ -130,7 +130,7 @@ namespace
             m_override_exclude_glass_materials = false;
             m_override_material = nullptr;
 
-            m_rendering_threads = 0;        // 0 = as many as there are logical cores
+            m_rendering_threads = -1;        // -1 = all available logical cores -1 are used for rendering
             m_enable_embree = false;
             m_low_priority_mode = true;
             m_use_max_procedural_maps = false;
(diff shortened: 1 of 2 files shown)


src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -369,7 +369,7 @@ void RendererSettings::apply_settings_to_interactive_config(asr::Project& projec
     params.insert_path("lighting_engine", "pt");
 
     if (m_rendering_threads == 0)
```
dictoon: I wonder if this whole block still makes sense, now that the default for m_rendering_threads is sane (-1). There's no special case for final configurations.

Mango-3: For ActiveShade mode I think it still makes sense that when rendering_threads == 0 (e.g. from loading saved scene files) it is always overridden to keep the UI responsive.  
For production rendering mode it is less of an issue as one does not actively manipulate the scene during rendering (as one usually does in ActiveShade mode).

dictoon: Ok.

In that case, the code that sets the number of rendering threads in `apply_common_settings()` could be moved to `apply_settings_to_final_config()`. I think that would make things a little clearer.

Mango-3: Yes, absolutely right.

src/appleseed-max-impl/appleseedrenderer/renderersettings.cpp
```diff
@@ -357,6 +353,7 @@ void RendererSettings::apply_settings_to_final_config(asr::Project& project) con
         params.insert_path("adaptive_tile_renderer.max_samples", m_adaptive_max_samples);
         params.insert_path("adaptive_tile_renderer.noise_threshold", m_adaptive_noise_threshold);
     }
+    params.insert_path("rendering_threads", m_rendering_threads);
```
dictoon: You need to preserve the logic that set the parameter to `auto` when it's 0 in Max.

Mango-3: Yes, definitely true.

PR #342, opened 2019-12-09
Fix interactive render crash on cancel

### LINKED ISSUE #341: Interactive rendering crashes Max when closing framebuffer window
Looks like a regression. I'll make a PR with fixes shortly.

usakhelo:
This should fix #341 
Reference to destructed local variable `search_paths` was kept in `MasterRenderer` class.

### DIFF
--- a/src/appleseed-max-impl/appleseedinteractive/interactivesession.h
+++ b/src/appleseed-max-impl/appleseedinteractive/interactivesession.h
@@ -45,6 +45,8 @@
 // Forward declarations.
 namespace renderer { class Camera; }
 namespace renderer { class Project; }
+namespace foundation { class SearchPaths; }
+
 class Bitmap;
 class IIRenderMgr;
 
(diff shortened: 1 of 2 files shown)


src/appleseed-max-impl/appleseedinteractive/interactivesession.h
```diff
@@ -43,8 +43,10 @@
 #include <thread>
 
 // Forward declarations.
-namespace renderer { class Camera; }
-namespace renderer { class Project; }
+namespace foundation { class SearchPaths; }
```
dictoon: A forward declaration is actually not enough in this case, you'll need to include `foundation/utility/searchpaths.h`.

usakhelo: I thought the same but it compiled and thought I still don't understand anything about c++ :)

dictoon: What most likely happens is that `SearchPaths` is already made available by an indirect `#include` statement, and the forward declaration is entirely redundant.

PR #345, opened 2020-01-04
Adapt to recent changes in renderer::TileCallback class

Mango-3:


src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.cpp
```diff
@@ -60,9 +60,13 @@ InteractiveTileCallback::InteractiveTileCallback(
...
 void InteractiveTileCallback::on_progressive_frame_update(
-    const asr::Frame*           frame)
+    const asr::Frame&           frame,
+    const double                time,
+    const std::uint64_t         samples,
+    const double                samples_per_pixel,
+    const std::uint64_t         samples_per_second)
 {
-    TileCallback::on_progressive_frame_update(frame);
+    TileCallback::on_progressive_frame_update(frame, time, samples, samples_per_pixel, samples_per_second);
```
usakhelo: My only comment: I would probably list arguments in a column:
```
TileCallback::on_progressive_frame_update(
    frame,
    time,
    samples,
    samples_per_pixel,
    samples_per_second);
```

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.h
```diff
@@ -56,7 +56,12 @@ class InteractiveTileCallback
         IIRenderMgr*                    iimanager,
```
dictoon: Unrelated to this PR but I couldn't help from noticing: `iimanager` doesn't make much sense as a name. I guess the first I is for "Interface" and the second is for "Interactive"? In that case maybe `irender_manager`? Or `render_manager`?

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.h
```diff
@@ -56,7 +56,12 @@ class InteractiveTileCallback
...
         renderer::IRendererController*  renderer_controller);
 
-    void on_progressive_frame_update(const renderer::Frame* frame) override;
+    void on_progressive_frame_update(
+        const renderer::Frame&          frame,
+        const double                    time,
+        const std::uint64_t             samples,
```
dictoon: Make sure to include `<cstdint>`.

src/appleseed-max-impl/appleseedinteractive/interactivetilecallback.h
Mango-3: True, The naming comes from the class interface declared in IIRenderManager.h, part of the Max SDK.
`This class represents the abstract (interface) for an interactive rendering
manager.`

PR #351, opened 2020-01-23
Improve appleseed-max object plugins support

dictoon:


usakhelo:
Hi @dictoon ,
I tried to compile and run this PR and it worked fine. Both in interactive and normal rendering.

PR #353, opened 2020-02-08
Collapse render settings panel rollouts - except Image Sampling

Mango-3:
- Collapses all rollouts (except the Image Sampling panel) of the render settings.
- Small changes to the load_tile() method following recent changes to appleseed-core.

### DIFF
--- a/src/appleseed-max-impl/utilities.cpp
+++ b/src/appleseed-max-impl/utilities.cpp
@@ -742,18 +742,11 @@ namespace
             return new MaxProceduralTextureSource(m_texmap, m_time);
         }
 
-        asf::Tile* load_tile(
+        asr::TilePtr load_tile(
             const size_t                tile_x,
             const size_t                tile_y) override
         {
-            return nullptr;
-        }
-
-        void unload_tile(
-            const size_t                tile_x,
-            const size_t                tile_y,
-            const asf::Tile*            tile) override
-        {
+            return asr::TilePtr::make_nullptr();
         }
 
       private:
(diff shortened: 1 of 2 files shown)


src/appleseed-max-impl/utilities.cpp
```diff
@@ -39,6 +39,7 @@
 #include "renderer/api/color.h"
 #include "renderer/api/source.h"
 #include "renderer/api/texture.h"
+#include "renderer/modeling/texture/tileptr.h"
```
dictoon: I'll make a PR on appleseed's side to include `renderer/modeling/texture/tileptr.h` into `renderer/api/texture.h`.

src/appleseed-max-impl/utilities.cpp
```diff
@@ -742,18 +743,12 @@ namespace
...
-            return nullptr;
-        }
-
-        void unload_tile(
-            const size_t                tile_x,
-            const size_t                tile_y,
-            const asf::Tile*            tile) override
-        {
+            asf::Tile* tile = nullptr;
+            return asr::TilePtr::make_non_owning(tile);
```
dictoon: I'll also add a `TilePtr::make_nullptr()` static method to make this cleaner.

src/appleseed-max-impl/utilities.cpp
Mango-3: True, should be part of the api, then adding '#include "renderer/modeling/texture/tileptr.h"'can be omitted.

src/appleseed-max-impl/utilities.cpp
Mango-3: Great, I was not sure if my implementation was the correct way to do it.

dictoon: Your implementation is perfectly correct.

